PR #18, opened 2023-10-05
[SASS-5838] Income Journey - generate scaffold, create new model generators for BigDecimals

Jake-Raffe:


app/models/HowMuchTradingAllowance.scala
harry-ritchie, line 1: Did the scaffold generate these model files? If they are specific to implementations of the tickets should they be done in the specific page tickets?

app/viewmodels/checkAnswers/AnyOtherIncomeSummary.scala
```diff
@@ -0,0 +1,44 @@
...
+import models.{CheckMode, UserAnswers}
+import pages.AnyOtherIncomePage
+import play.api.i18n.Messages
+import uk.gov.hmrc.govukfrontend.views.viewmodels.summarylist.SummaryListRow
+import viewmodels.govuk.summarylist._
+import viewmodels.implicits._
+
+object AnyOtherIncomeSummary {
+
+  def row(answers: UserAnswers, taxYear: Int)(implicit messages: Messages): Option[SummaryListRow] =
```
harry-ritchie, line 29: Me and Hassan were looking at the way we generate the `SummaryListRow` in the CYA view models, and we were thinking it would make sense to handle the `None` case (where no user answers are present for the `AnyOtherIncomePage`) inside this method, instead of the exception being generated in the controller itself. See [here](https://github.com/hmrc/income-tax-self-employment-frontend/pull/16/files#diff-304f6b650510e6effffaaf23b9125050da29d49161aa983538b5839909920f26R29-R44) 

What are your thoughts?

app/models/HowMuchTradingAllowance.scala
Jake-Raffe: Generated by the scaffold and for now will be kept separate from the Page models

edem-nsefik-hmrc: With scaffold page development is done across multiple pages of a journey and tickets are done for each of the multi page activities, such as page generation, navigation, content etc

harry-ritchie (dismissed):

harry-ritchie (approved):

app/viewmodels/checkAnswers/AnyOtherIncomeSummary.scala
Jake-Raffe: Some summary rows need to be optional and the user answers from most pages will arrive not in an Option after ticket SASS-5841 is implemented.
For that reason the summary models can be left as is or can be edited on a case by case bases in the way you have described when ticket SASS-5842 is being done. I shall leave a link to this comment in ticket 5842.

PR #31, opened 2023-10-24
[SASS-5840 & SASS-5841] Add navigation and get/set UserAnswers to all Income journey pages

Jake-Raffe:


app/controllers/journeys/income/TradingAllowanceAmountController.scala
```diff
@@ -42,38 +42,38 @@ class TradingAllowanceAmountController @Inject() (override val messagesApi: Mess
...
-      }
-
-      Ok(view(preparedForm, mode, authUserType(request.user.isAgent), taxYear))
-  }
-
-  def onSubmit(taxYear: Int, mode: Mode): Action[AnyContent] = (identify andThen getData) async { // TODO add requireData SASS-5841
-    implicit request =>
-      val turnoverAmount: BigDecimal = {
-        val turnover: BigDecimal = request.userAnswers.getOrElse(UserAnswers(request.userId)).get(TurnoverIncomeAmountPage).getOrElse(1000.00)
+        val turnover: BigDecimal = request.userAnswers.get(TurnoverIncomeAmountPage, Some(businessId)).getOrElse(1000.00)
```
lukastymo, line 48: It would be nice to keep controllers logic-free. 

- Any ifs and magic numbers ideally could live in models (if we don't return Futures) or in service layer (if we do).
- Also, can you extract 1000 to const with a name so it's clearer what it does

app/controllers/journeys/income/TradingAllowanceAmountController.scala
```diff
@@ -42,38 +42,38 @@ class TradingAllowanceAmountController @Inject() (override val messagesApi: Mess
...
               _              <- sessionRepository.set(updatedAnswers)
-            } yield Redirect(navigator.nextPage(TradingAllowanceAmountPage, mode, updatedAnswers, taxYear))
+            } yield Redirect(navigator.nextPage(TradingAllowanceAmountPage, mode, updatedAnswers, taxYear, businessId))
         )
   }
 
+  def onPageLoad(taxYear: Int, businessId: String, mode: Mode): Action[AnyContent] = (identify andThen getData andThen requireData) {
+    implicit request =>
+      val turnoverAmount: BigDecimal = {
+        val turnover: BigDecimal = request.userAnswers.get(TurnoverIncomeAmountPage, Some(businessId)).getOrElse(1000.00)
```
lukastymo, line 66: similar here, can we move logic out of controller?

app/controllers/journeys/income/TradingAllowanceAmountController.scala
Jake-Raffe: will make this change

lukastymo (dismissed):

app/controllers/journeys/SectionCompletedStateController.scala
```diff
@@ -35,7 +35,7 @@ import scala.concurrent.{ExecutionContext, Future}
 
 class SectionCompletedStateController @Inject()(override val messagesApi: MessagesApi,
                                                 selfEmploymentConnector: SelfEmploymentConnector,
-                                                navigator: Navigator,
+                                                navigator: IncomeNavigator,
```
harry-ritchie, line 38: Does it make sense for the section completed controller to use the income navigator? Would it be better to have a generic navigator for pages not tied to a specific journey (although there is just the one page at the moment)?

app/controllers/journeys/abroad/SelfEmploymentAbroadCYAController.scala
```diff
@@ -50,9 +50,9 @@ class SelfEmploymentAbroadCYAController @Inject()(override val messagesApi: Mess
     Ok(view(taxYear, summaryList, nextRoute, isAgent))
   }
 
-  private def nextPageUrl(taxYear: Int, optBusinessId: String, navigator: Navigator)(implicit request: DataRequest[AnyContent]): String =
+  private def nextPageUrl(taxYear: Int, optBusinessId: String, navigator: AbroadNavigator)(implicit request: DataRequest[AnyContent]): String =
```
harry-ritchie, line 53: BusinessID is no longer optional in the navigator, this should be renamed from `optBusinessId` to `businessId`

app/controllers/journeys/income/OtherIncomeAmountController.scala
```diff
@@ -45,34 +45,32 @@ class OtherIncomeAmountController @Inject() (override val messagesApi: MessagesA
...
+              formWithErrors => Future.successful(BadRequest(view(formWithErrors, mode, authUserType(request.user.isAgent), taxYear, businessId))),
               value =>
                 for {
-                  updatedAnswers <- Future.fromTry(request.userAnswers.getOrElse(UserAnswers(request.userId)).set(OtherIncomeAmountPage, value))
+                  updatedAnswers <- Future.fromTry(request.userAnswers.set(OtherIncomeAmountPage, value, Some(businessId)))
                   _              <- sessionRepository.set(updatedAnswers)
                 } yield Redirect(
-                  navigator.nextPage(OtherIncomeAmountPage, mode, updatedAnswers, taxYear)
-                ) // TODO 5840 use 'accountingType.equals("ACCRUAL")' in this .nextPage method
+                  navigator.nextPage(OtherIncomeAmountPage, mode, updatedAnswers, taxYear, businessId, Some(accountingType.equals("ACCRUAL")))
```
harry-ritchie, line 72: Might be worth pulling out `accountingType.equals("ACCRUAL"))` into a `isAccrual` (or wtte) variable and passing in `Some(isAccrual)` to the method, just for readability.

app/controllers/journeys/income/TradingAllowanceAmountController.scala
```diff
@@ -42,38 +42,38 @@ class TradingAllowanceAmountController @Inject() (override val messagesApi: Mess
...
             for {
-              updatedAnswers <- Future.fromTry(request.userAnswers.getOrElse(UserAnswers(request.userId)).set(TradingAllowanceAmountPage, value))
+              updatedAnswers <- Future.fromTry(request.userAnswers.set(TradingAllowanceAmountPage, value, Some(businessId)))
               _              <- sessionRepository.set(updatedAnswers)
-            } yield Redirect(navigator.nextPage(TradingAllowanceAmountPage, mode, updatedAnswers, taxYear))
+            } yield Redirect(navigator.nextPage(TradingAllowanceAmountPage, mode, updatedAnswers, taxYear, businessId))
         )
   }
 
+  def onPageLoad(taxYear: Int, businessId: String, mode: Mode): Action[AnyContent] = (identify andThen getData andThen requireData) {
```
harry-ritchie, line 62: Can we put `onPageLoad` at the top of the class, i.e. before the `onSubmit` method, to align with what we do in other controllers.

app/navigation/IncomeNavigator.scala
```diff
@@ -0,0 +1,122 @@
...
+
+    case TurnoverIncomeAmountPage => _ => (taxYear, businessId, _) => AnyOtherIncomeController.onPageLoad(taxYear, businessId, NormalMode)
+
+    case AnyOtherIncomePage =>
+      userAnswers =>
+        (taxYear, businessId, optIsAccrual) =>
+          userAnswers.get(AnyOtherIncomePage, Some(businessId)) match {
+            case Some(true)                                   => OtherIncomeAmountController.onPageLoad(taxYear, businessId, NormalMode)
+            case Some(false) if optIsAccrual.getOrElse(false) => TurnoverNotTaxableController.onPageLoad(taxYear, businessId, NormalMode)
+            case Some(false) if !optIsAccrual.getOrElse(true) => TradingAllowanceController.onPageLoad(taxYear, businessId, NormalMode)
```
harry-ritchie, line 56: Might be a personal preference thing, but I would prefer using`.contains` instead of `.getOrElse` as the latter implies we want to provide a default value for a specific requirement, but i think it is just used here to handle the None case for compile reasons. 

I would prefer:

`case Some(false) if optIsAccrual.contains(true) => ...
  case Some(false) if optIsAccrual.contains(false) => ...`

I think it also helps with readability.

test/controllers/journeys/income/AnyOtherIncomeControllerSpec.scala
harry-ritchie, line 1: Do we still want the tests starting on line 123 and 317 to still be ignored? As we do require data now.

test/controllers/journeys/income/AnyOtherIncomeControllerSpec.scala
```diff
@@ -17,40 +17,52 @@
...
+  val formProvider           = new AnyOtherIncomeFormProvider()
+  val businessId             = "SJPR05893938418"
+  val otherIncomeAmountCall  = OtherIncomeAmountController.onPageLoad(taxYear, businessId, NormalMode)
+  val turnoverNotTaxableCall = TurnoverNotTaxableController.onPageLoad(taxYear, businessId, NormalMode)
+  val tradingAllowanceCall   = TradingAllowanceController.onPageLoad(taxYear, businessId, NormalMode)
+
+  val mockService: SelfEmploymentService       = mock[SelfEmploymentService]
+  val mockSessionRepository: SessionRepository = mock[SessionRepository]
+
+  val answerIsTrueOnwardRoute = otherIncomeAmountCall
```
harry-ritchie, line 57: Probably don't need this val, I don't think it adds anything to the test other than complexity.

test/controllers/journeys/income/AnyOtherIncomeControllerSpec.scala
```diff
@@ -17,40 +17,52 @@
...
 
-  val formProvider       = new AnyOtherIncomeFormProvider()
-  val formWithIndividual = formProvider("individual")
-  val formWithAgent      = formProvider("agent")
+  def answerIsFalseOnwardRoute(isAccrual: Boolean) = if (isAccrual) turnoverNotTaxableCall else tradingAllowanceCall
```
harry-ritchie, line 59: Bit more readable to pass in the accounting type to this method and pattern match on its value, to avoid passing in boolean values. 

I think the method name could be improved, e.g. `redirectCallFromNoResponse` or wtte, so this could all become:

`def redirectCallFromNoResponse(accountingType: String) = accountingType match {
       case "ACCRUAL" => turnoverNotTaxableCall
       case "CASH"        => tradingAllowanceCall
}`

test/controllers/journeys/income/AnyOtherIncomeControllerSpec.scala
```diff
@@ -127,40 +139,131 @@ class AnyOtherIncomeControllerSpec extends SpecBase with MockitoSugar {
...
 
-          val result = route(application, request).value
+            val result = route(application, request).value
 
-          status(result) mustEqual SEE_OTHER
-          redirectLocation(result).value mustEqual onwardRoute.url
+            status(result) mustEqual SEE_OTHER
+            redirectLocation(result).value mustEqual otherIncomeAmountCall.url
+          }
+        }
```
harry-ritchie, line 168: I think the tests you've written here (for checking that a certain redirection happens based upon what you've set the user answers to and what the accounting type is) are not actually testing what you'd hope them to. 

The reason for this is that the responsibility of redirection based upon user answers and the accounting type in the controller is delegated to the navigator. In the tests you've written, we are essentially mocking our navigator, and so regardless of what we set our user answers or the accounting type to, this has no effect. 

To highlight what I mean, if in this test we change the user answers to false and the accounting type to "CASH", then the tests still pass. I think what you are trying to test is the integration of the controller with the navigator, which is are fair tests to do (maybe we should be doing this???), however these tests should be unit tests. 

I think for our unit tests, we could do the following:

- Test that when `selfEmploymentService.getAccountingType` returns Left(x) (which we can achieve by mocking like you have already done), that we redirect to `JourneyRecoveryController`

- Test that when it returns an accounting type, and our form data does not match the format of the form provider (which we already have tests for), we pass in that form with errors to the view

- Test that when it returns an accounting type, and our form data matches the form provider format, that user answers are set correctly (can be done without mocking) and that the session repository has successfully set our answers (done via mocking), that we are passing those updated user answers and the accounting type to the navigator (this can be done by setting the expectations on the params during mocking in Mockito, i.e. `(eq(updatedUserAnswers))` ) and we redirect to it's return value.

Hope this makes sense. Happy to discuss it if i have misunderstood.

Note - I think this comment also applies to the other controller tests where you've done this sort of testing.

test/controllers/journeys/income/HowMuchTradingAllowanceControllerSpec.scala
harry-ritchie, line 1: We still have some ignored tests here for testing that we have enabled `requireData`, can these be unignored now?

test/controllers/journeys/income/IncomeNotCountedAsTurnoverControllerSpec.scala
```diff
@@ -113,7 +113,37 @@ class IncomeNotCountedAsTurnoverControllerSpec extends SpecBase with MockitoSuga
 
     "onSubmit" - {
 
-      "must redirect to the next page when valid data is submitted" in {
+      "must redirect to the Non-Turnover Income Amount page when a user answer 'Yes' is submitted" in {
+
+        val userAnswer = true
+
+        val mockSessionRepository = mock[SessionRepository]
```
harry-ritchie, line 120: This is used in a couple places so could we pull it to the top of the class?

test/controllers/journeys/income/NonTurnoverIncomeAmountControllerSpec.scala
harry-ritchie, line 1: Again, there are ignored tests for require data.

test/navigation/IncomeNavigatorSpec.scala
harry-ritchie, line 1: I think we are missing UTs for when the page is `OtherIncomeAmountPage`

app/controllers/journeys/income/OtherIncomeAmountController.scala
harry-ritchie: On second thoughts - this is not as significant as I previously thought. I'll leave it up to you.

test/navigation/IncomeNavigatorSpec.scala
Jake-Raffe: Good point, thank you

test/controllers/journeys/income/AnyOtherIncomeControllerSpec.scala
Jake-Raffe: We will discuss this with the team to decide what's best going forwards

harry-ritchie (approved):

PR #33, opened 2023-10-25
[SASS-5963] Create BigDecimal to String formatter

Jake-Raffe:


app/services/SelfEmploymentService.scala
```diff
@@ -51,4 +53,11 @@ class SelfEmploymentService @Inject() (connector: SelfEmploymentConnector)(impli
     if (turnover > maxIncomeTradingAllowance) maxIncomeTradingAllowance else turnover
   }
 
+  def convertBigDecimalToMoneyString(amount: BigDecimal): String = {
```
lukastymo, line 56: we can even move it to service companion object so it is a simple function as it does not have any dependencies to the service fields.

test/services/SelfEmploymentServiceSpec.scala
```diff
@@ -114,13 +114,23 @@ class SelfEmploymentServiceSpec extends SpecBase with MockitoSugar {
...
 
         service.getIncomeTradingAllowance(businessId, userAnswersLargeTurnover) mustEqual maxIncomeTradingAllowance
         service.getIncomeTradingAllowance(businessId, userAnswersEqualToMax) mustEqual maxIncomeTradingAllowance
       }
     }
   }
 
+  "convertBigDecimalToMoneyString" - {
+    "should format BigDecimals to String with commas every thousand, and to two decimal places unless a whole number" in {
+      val bigDecimalSeq: Seq[BigDecimal]  = Seq(1000000000, 1000.00, 1000.1, 1000.10, 1000.01)
```
lukastymo, line 129: what about case with < 1000, so we don't have a comma

lukastymo (dismissed):

lukastymo:
LGTM, few minor comments

app/services/SelfEmploymentService.scala
```diff
@@ -51,4 +53,11 @@ class SelfEmploymentService @Inject() (connector: SelfEmploymentConnector)(impli
...
+    val numberFormat  = NumberFormat.getNumberInstance(Locale.US)
```
lukastymo, line 57: hmm, are you sure it should be US? that's what we use in other places? Not Locale.UK?

lukastymo (approved):

PR #35, opened 2023-10-26
[SASS-5975] Create content for the expenses pages 4-6

Jake-Raffe:


app/assets/stylesheets/application.scss
```diff
@@ -13,6 +13,14 @@ strong {
   white-space: nowrap;
 }
 
+.no-padding-top {
+  padding-top: 0;
+}
+
```
Jake-Raffe, line 19: These are placeholders and will be replaced in [SASS-5949](https://jira.tools.tax.service.gov.uk/browse/SASS-5949)

app/viewmodels/ContentStringViewModel.scala
```diff
@@ -0,0 +1,36 @@
...
+                                       optAuthUserType: Option[String] = None,
+                                       headingExtraClasses: String = "",
+                                       hintExtraClasses: String = "")(implicit messages: Messages): String = {
+    val optAuthUserString = if (optAuthUserType.isEmpty) "" else s".${optAuthUserType.get}"
+    val heading           = messages(s"${page.toString}.title$optAuthUserString")
+    val hint              = messages(s"${page.toString}.hint$optAuthUserString")
+
+    s"<div> <legend class='govuk-heading-l govuk-!-margin-0 $headingExtraClasses'> $heading </legend>" +
+      s"<div class='govuk-hint $hintExtraClasses'> $hint </div> </div>"
+  }
```
Jake-Raffe, line 34: This is necessary to have the error formatting display correctly. The red error bar must encompass the heading and the hint, meaning both must be passed as `Content` in the RadioButton's legend. If you know a better way of doing this then let me know but this was the best solution I could find.
I will attach a comment to the ticket that shows the difference in browser.

app/viewmodels/ContentStringViewModel.scala
```diff
@@ -0,0 +1,36 @@
...
+import pages.Page
+import play.api.i18n.Messages
+
+object ContentStringViewModel {
+
+  def buildLegendHeadingWithHintString(page: Page,
+                                       optAuthUserType: Option[String] = None,
+                                       headingExtraClasses: String = "",
+                                       hintExtraClasses: String = "")(implicit messages: Messages): String = {
+    val optAuthUserString = if (optAuthUserType.isEmpty) "" else s".${optAuthUserType.get}"
```
lukastymo, line 28: or we can avoid if: ` optAuthUserType.map(userType => s".$userType").getOrElse("")` ?

lukastymo (approved):

app/viewmodels/ContentStringViewModel.scala
Jake-Raffe: will make this change

PR #41, opened 2023-10-30
SASS-6007: Add model for all income journey answers 

harry-ritchie:


app/models/journeys/income/IncomeJourneyAnswers.scala
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.journeys.income
+
+import models.{HowMuchTradingAllowance, TradingAllowance}
+import play.api.libs.json.{Json, OFormat}
+
+case class IncomeJourneyAnswers(incomeNotCountedAsTurnover: Boolean,
```
lukastymo, line 22: it is fine, just one observation we may represent incorrect states:

incomeNotCountedAsTurnover=true
nonTurnoverIncomeAmount=None

ideally we should aim in not possible to represent incorrect states.

It's fine now, we can refine closer when we add validation

lukastymo (dismissed):

app/models/journeys/income/IncomeJourneyAnswers.scala
harry-ritchie: That's a good point. I'll have a look at putting this constraint in this PR.

harry-ritchie: We have agreed this will now be done in the backend

lukastymo (approved):

PR #43, opened 2023-10-31
[SASS-6018] IncomeJ - Clear old data when a user changes their answer

Jake-Raffe:


app/controllers/journeys/income/AnyOtherIncomeController.scala
```diff
@@ -67,8 +67,13 @@ class AnyOtherIncomeController @Inject() (override val messagesApi: MessagesApi,
               formWithErrors => Future.successful(BadRequest(view(formWithErrors, mode, userType(request.user.isAgent), taxYear, businessId))),
               value =>
                 for {
-                  updatedAnswers <- Future.fromTry(request.userAnswers.set(AnyOtherIncomePage, value, Some(businessId)))
-                  _              <- sessionRepository.set(updatedAnswers)
+                  updatedAnswers <- Future.fromTry {
+                    val userAnswers =
+                      if (!value) request.userAnswers.remove(OtherIncomeAmountPage, Some(businessId)).get
```
lukastymo, line 72: minor: 

- we can try to generalize it a bit, it could live in some model to keep our controllers logic 
- I would try to avoid as much as possible throwing exceptions (.get) can cause an exception. We can live it for now because I don't know if we have some well-defined Error Handling, and well defined Left part so we could return Left in Either for error scenarios

lukastymo (dismissed):

app/controllers/journeys/income/TradingAllowanceController.scala
```diff
@@ -72,8 +73,18 @@ class TradingAllowanceController @Inject() (override val messagesApi: MessagesAp
...
-                  updatedAnswers <- Future.fromTry(request.userAnswers.set(TradingAllowancePage, value, Some(businessId)))
-                  _              <- sessionRepository.set(updatedAnswers)
+                  updatedAnswers <- Future.fromTry {
+                    val userAnswers =
+                      if (value.equals(DeclareExpenses))
+                        request.userAnswers
+                          .remove(HowMuchTradingAllowancePage, Some(businessId))
+                          .get
+                          .remove(TradingAllowanceAmountPage, Some(businessId))
+                          .get
```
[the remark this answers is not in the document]
Jake-Raffe: agreed to handle in a future ticket

app/controllers/journeys/income/AnyOtherIncomeController.scala
Jake-Raffe: agreed to handle in a future ticket

lukastymo (approved):

PR #47, opened 2023-11-01
[SASS-6024] Expenses 3/3 - Add content to pages 1/2

Jake-Raffe:


app/controllers/journeys/expenses/FinancialExpensesController.scala
```diff
@@ -42,28 +48,50 @@ class FinancialExpensesController @Inject() (override val messagesApi: MessagesA
...
-    }
+  def onPageLoad(mode: Mode): Action[AnyContent] = (identify andThen getData) async { implicit request =>
+    selfEmploymentService.getAccountingType(request.user.nino, businessId, request.user.mtditid) map {
+      case Left(_) => Redirect(JourneyRecoveryController.onPageLoad())
+      case Right(accountingType) =>
+        val preparedForm = request.userAnswers.getOrElse(UserAnswers(request.userId)).get(FinancialExpensesPage, Some(businessId)) match {
+          case None        => formProvider(userType(request.user.isAgent))
+          case Some(value) => formProvider(userType(request.user.isAgent)).fill(value)
+        }
+        val legendContentString = buildLegendHeadingWithHintString(
```
harry-ritchie, line 62: If both `legendContentString` used in `onSubmit` and here are going to always be the same, it could be worth pulling them out into a method to reduce duplication.

Also, we don't need the suffix `..String` on the end, the IDE tells us this.

Thirdly (if this is the case - ignore part 1), should the controller be doing this? Or is it more of the views responsibility to be doing this than the controller?

app/controllers/journeys/expenses/WorkFromBusinessPremisesController.scala
```diff
@@ -54,18 +54,20 @@ class WorkFromBusinessPremisesController @Inject() (override val messagesApi: Me
...
+      s"workFromBusinessPremises.title.${userType(request.user.isAgent)}",
+      s"workFromBusinessPremises.hint.${userType(request.user.isAgent)}",
+      headingClasses = "govuk-fieldset__legend govuk-fieldset__legend--l"
+    )
 
     Ok(view(preparedForm, mode, userType(request.user.isAgent), taxYear, businessId, radioContentString))
   }
 
   def onSubmit(mode: Mode): Action[AnyContent] = (identify andThen getData) async { implicit request =>
     val radioContentString = buildLegendHeadingWithHintString(
```
harry-ritchie, line 66: All comments with regards to `legendContentString` in `FinancialExpensesController` also apply here.

conf/messages.en
```diff
@@ -18,6 +18,7 @@ site.canInclude.individual = You can include:
 site.canInclude.agent = Your client can include:
 site.cannotInclude.individual = You cannot include:
 site.cannotInclude.agent = Your client cannot include:
+site.selectAllTheApply = Select all that apply.
```
harry-ritchie, line 21: Should we have this named as `site.selectAllThatApply` instead?

conf/messages.cy
```diff
@@ -18,6 +18,7 @@ site.canInclude.individual = You can include:
 site.canInclude.agent = Your client can include:
 site.cannotInclude.individual = You cannot include:
 site.cannotInclude.agent = Your client cannot include:
+site.selectAllTheApply = Select all that apply.
```
harry-ritchie, line 21: `site.selectAllThatApply`?

app/controllers/journeys/expenses/FinancialExpensesController.scala
Jake-Raffe: Will remove string and separate it from the methods, thanks. I found this was the only way to pass both parts of the content into the the view's LegendViewModel. Had to do the same thing for the task list.

conf/messages.en
Jake-Raffe: thanks, typo

app/controllers/journeys/expenses/WorkFromBusinessPremisesController.scala
harry-ritchie: @Jake-Raffe Could we also do the same with pulling the content method out into a shared method here like we did in financial expenses controller?

Jake-Raffe: Thanks, missed this one

harry-ritchie (approved):

PR #65, opened 2023-11-10
SASS-6110: Adding content to Office Supplies journey screens, including Navigation 

harry-ritchie:


app/navigation/OfficeSuppliesNavigator.scala
```diff
@@ -0,0 +1,53 @@
...
+
+package navigation
+
+import controllers.journeys.expenses.officeSupplies.routes.{OfficeSuppliesCYAController, OfficeSuppliesDisallowableAmountController}
+import controllers.standard.routes.JourneyRecoveryController
+import models.database.UserAnswers
+import models.{CheckMode, Mode, NormalMode}
+import pages.Page
+import pages.expenses.officeSupplies.{OfficeSuppliesAmountPage, OfficeSuppliesDisallowableAmountPage}
+import play.api.mvc.Call
```
Jake-Raffe, line 25: did we decide we were going to have every subjourney in it's own navigator or that we'd put all non-tailoring expenses in the expensesNavigator?

app/views/journeys/expenses/officeSupplies/OfficeSuppliesDisallowableAmountView.scala.html
```diff
@@ -16,32 +16,37 @@
...
             @govukErrorSummary(ErrorSummaryViewModel(form))
         }
 
+        @heading(messages(s"officeSuppliesDisallowableAmount.title.$authUserType", allowableAmount.toString), None, extraClasses = "govuk-!-margin-bottom-3")
+
         @govukInput(
             InputViewModel(
                 field = form("value"),
-                label = LabelViewModel(messages("officeSuppliesDisallowableAmount.heading")).asPageHeading()
+                label = LabelViewModel("").asPageHeading()
```
Jake-Raffe, line 45: if there's no label it doesn't need .asPageHeading(). But the error should encompass the title on this page so think the heading needs to be the label, not in the @heading() on line 40

app/navigation/OfficeSuppliesNavigator.scala
harry-ritchie: No, we didn't decide to do it this way. I just had to do something then and there. What do you think is best?

harry-ritchie: I think if we have all the expenses in one navigator, the file will be quite large. Maybe we should chat to @lukastymo - what do you think?

test/navigation/OfficeSuppliesNavigatorSpec.scala
```diff
@@ -0,0 +1,76 @@
...
+import pages.expenses.officeSupplies.{OfficeSuppliesAmountPage, OfficeSuppliesDisallowableAmountPage}
+
+class OfficeSuppliesNavigatorSpec extends AnyWordSpec with Matchers {
+
+  private val navigator = new OfficeSuppliesNavigator
+
+  private val taxYear     = 2024
+  private val userAnswers = UserAnswers("some_id")
+  private val businessId  = "some_businessId"
+
```
Jake-Raffe, line 35: taxYear, userAnswersId, emptyUserAnswers and businessId exist in the SpecBase if you want to extend that. CheckMode and NormalMode can be used without the vals

test/viewmodels/checkAnswers/expenses/officeSupplies/OfficeSuppliesDisallowableAmountSummarySpec.scala
```diff
@@ -26,7 +26,9 @@ import uk.gov.hmrc.govukfrontend.views.viewmodels.summarylist.SummaryListRow
 
 class OfficeSuppliesDisallowableAmountSummarySpec extends AnyWordSpec with Matchers {
```
Jake-Raffe, line 27: these are extended in SpecBase already, think we should keep consistent with the scaffold and all other Specs

app/views/journeys/expenses/officeSupplies/OfficeSuppliesDisallowableAmountView.scala.html
harry-ritchie: Good spot.

test/navigation/OfficeSuppliesNavigatorSpec.scala
harry-ritchie: The reason I didn't as I find AnyWordSpec more readable, but SpecBase uses AnyFreeSpec. I can use SpecBase to be consistent with the other files.

Jake-Raffe (dismissed):

Jake-Raffe (approved):

PR #68, opened 2023-11-13
WIP SASS-6114: Scaffold Repairs and Maintenance

lukastymo:


test/controllers/journeys/expenses/repairsandmaintenance/RepairsAndMaintenanceAmountControllerSpec.scala
```diff
@@ -0,0 +1,43 @@
...
+import play.api.mvc.Request
+import play.twirl.api.{Html, HtmlFormat}
+import views.html.journeys.expenses.repairsandmaintenance.RepairsAndMaintenanceAmountView
+
+class RepairsAndMaintenanceAmountControllerSpec
+    extends BigDecimalGetAndPostQuestionBaseSpec("RepairsAndMaintenanceAmountController", RepairsAndMaintenanceAmountPage) {
+  lazy val onPageLoadRoute = RepairsAndMaintenanceAmountController.onPageLoad(taxYear, stubbedBusinessId, NormalMode).url
+  lazy val onSubmitRoute   = RepairsAndMaintenanceAmountController.onSubmit(taxYear, stubbedBusinessId, NormalMode).url
+
+  def createForm(userType: UserType): Form[BigDecimal] = new RepairsAndMaintenanceAmountFormProvider()(userType)
```
harry-ritchie, line 38: Could we add `override` to these methods to be clear they are overidden?

test/base/ControllerSpec.scala
```diff
@@ -0,0 +1,70 @@
...
+    when(mockSessionRepository.set(any())) thenReturn Future.successful(true)
+
+  }
+
+  def onwardRoute = Call("GET", "/foo")
+
+  private def createApp(userType: UserType, answers: Option[UserAnswers], mockSessionRepository: SessionRepository) = {
+    applicationBuilder(answers, isAgent(userType.toString))
+      .overrides(
+        bind[ExpensesNavigator].toInstance(new FakeExpensesNavigator(onwardRoute)),
```
harry-ritchie, line 64: Is there some way we could make this not specific to Expenses so this trait can be used for all journeys?

test/base/ControllerSpec.scala
```diff
@@ -0,0 +1,70 @@
...
+    (Language.Welsh, UserType.Agent)
+  )
+
+  abstract class TestApp(userType: UserType, answers: Option[UserAnswers]) {
+    private val mockSessionRepository = mock[SessionRepository]
+
+    implicit val application              = createApp(userType, answers, mockSessionRepository)
+    implicit val messagesApi: MessagesApi = application.injector.instanceOf[MessagesApi]
+
+    private val onwardRoute = Call("GET", "/foo")
```
harry-ritchie, line 53: `onwardRoute` is not used anywhere.

test/base/ControllerSpec.scala
```diff
@@ -0,0 +1,70 @@
...
+import play.api.inject.bind
+
+import scala.concurrent.Future
+import scala.reflect.ClassTag
+
+trait ControllerSpec extends SpecBase with MockitoSugar with TableDrivenPropertyChecks {
+
+  val langUserTypeCases = Table(
+    ("Language", "userType"),
+    (Language.English, UserType.Individual: UserType),
```
harry-ritchie, line 41: Don't need the `:UserType here`

test/base/BigDecimalGetAndPostQuestionBaseSpec.scala
```diff
@@ -0,0 +1,120 @@
...
+            status(result) mustEqual OK
+            contentAsString(result) mustEqual renderView(getRequest, messages(application), application)(
+              form,
+              NormalMode,
+              taxYear,
+              stubbedBusinessId).toString
+          }
+        }
+
+        "Return OK for a GET if previous answer exists" in new TestApp(userType, Some(filledUserAnswers)) {
```
harry-ritchie, line 71: Maybe rephrase to `Return OK for a GET if a previous answer exists, with the view populated with the previous answer`?

test/base/BigDecimalGetAndPostQuestionBaseSpec.scala
```diff
@@ -0,0 +1,120 @@
...
+
+  def getRequest  = FakeRequest(GET, onPageLoadRoute)
+  def postRequest = FakeRequest(POST, onSubmitRoute).withFormUrlEncodedBody(("value", validAnswer.toString))
+
+  forAll(langUserTypeCases) { case (lang, userType) =>
+    s"$controllerName for ${lang.entryName} and ${userType}" - {
+      val form: Form[BigDecimal] = createForm(userType)
+
+      "Loading page" - {
+        "Redirect to Journey Recover for a GET if existing data not found " in new TestApp(userType, None) {
```
harry-ritchie, line 51: Maybe rephrase as `Redirect to Journey Recover for a GET if **prerequisite** data not found`, as the wording to me is very similar to the test below.

test/base/BigDecimalGetAndPostQuestionBaseSpec.scala
```diff
@@ -0,0 +1,120 @@
...
+              NormalMode,
+              taxYear,
+              stubbedBusinessId).toString
+          }
+
+        }
+      }
+
+      "Submitting page" - {
+        "Redirect to Journey Recovery for POST if no existing data is found" in new TestApp(userType, None) {
```
harry-ritchie, line 86: `..no prerequisite data is found`?

test/base/BigDecimalGetAndPostQuestionBaseSpec.scala
```diff
@@ -0,0 +1,120 @@
...
+import play.api.test.Helpers._
+import play.twirl.api.HtmlFormat
+
+abstract case class BigDecimalGetAndPostQuestionBaseSpec(controllerName: String, page: QuestionPage[BigDecimal]) extends ControllerSpec {
+  val onPageLoadRoute: String
+  val onSubmitRoute: String
+
+  def createForm(userType: UserType): Form[BigDecimal]
+
+  def renderView(implicit request: Request[_], messages: Messages, application: Application): (Form[_], Mode, Int, String) => HtmlFormat.Appendable
```
harry-ritchie, line 38: If we were to use this method for other BigDecimal question pages, the inputs to the views may slightly change, e.g. one may require accountingType passed into the view, and another may not (this is also the case with the form providers). What would be the best way to accommodate for this in the base class?

test/controllers/journeys/expenses/repairsandmaintenance/RepairsAndMaintenanceAmountControllerSpec.scala
lukastymo: the rule I used before about override in Scala was:
- do not use it if you add implementation (implement trait)
- add override if there is some implementation already and you really change, override behavior.

Here we don't change any implementation so override is omitted.

Let me know if you like it.

test/base/ControllerSpec.scala
lukastymo: used in BigDecimalGetAndPostQuestionBaseSpec - I think we mock navigation and test if we redirect to this route - it may be common for all ControllerSpec

test/base/ControllerSpec.scala
lukastymo: done

test/base/BigDecimalGetAndPostQuestionBaseSpec.scala
lukastymo: it's a bit long and too descriptive, but I can do it to be precise what the test does

test/base/BigDecimalGetAndPostQuestionBaseSpec.scala
lukastymo: done

test/base/BigDecimalGetAndPostQuestionBaseSpec.scala
lukastymo: done

test/controllers/journeys/expenses/repairsandmaintenance/RepairsAndMaintenanceAmountControllerSpec.scala
harry-ritchie: Makes sense!

test/base/ControllerSpec.scala
lukastymo: will do

test/base/BigDecimalGetAndPostQuestionBaseSpec.scala
lukastymo: ok, made a fix to accommodate accountingType and other more generic Types. We can pass now TestScenario

harry-ritchie (approved):

PR #83, opened 2023-11-17
[SASS-5631] Add expenses to tasklist

Jake-Raffe:


app/viewmodels/TradeJourneyStatusesViewModel.scala
```diff
@@ -43,9 +44,13 @@ object TradeJourneyStatusesViewModel {
       if (optJourney.isEmpty) None else optJourney.head.completedState
     }
```
lukastymo, line 45: Let's simplify this buildSummaryList, as a started let's move getStatus out of the buildSummaryList and make it norma def getStatus

app/viewmodels/TradeJourneyStatusesViewModel.scala
```diff
@@ -17,10 +17,11 @@
 package viewmodels
 
 import controllers.journeys.abroad.routes.{SelfEmploymentAbroadCYAController, SelfEmploymentAbroadController}
+import controllers.journeys.expenses.tailoring.routes.OfficeSuppliesController
 import controllers.journeys.income.routes.{IncomeCYAController, IncomeNotCountedAsTurnoverController}
 import models._
 import models.journeys.Journey
-import models.journeys.Journey.{Abroad, Income}
+import models.journeys.Journey.{Abroad, ExpensesTailoring, Income}
```
lukastymo, line 24: private val completedStatus      = "completed"
  private val inProgressStatus     = "inProgress"
  private val notStartedStatus     = "notStarted"
  private val cannotStartYetStatus = "cannotStartYet"

These are important statuses, let's give them proper types

```
import enumeratum._

sealed trait TradeJourneyStatus extends EnumEntry {
  override def entryName: String = toString.toLowercase // if you need them lowercased
}

object TradeJourneyStatus extends Enum[TradeJourneyStatus] {
  val values = findValues

  case object Completed extends AccountingType
  case object InProgress    extends AccountingType
...
}
```

app/viewmodels/TradeJourneyStatusesViewModel.scala
```diff
@@ -43,9 +44,13 @@ object TradeJourneyStatusesViewModel {
       if (optJourney.isEmpty) None else optJourney.head.completedState
     }
 
-    val (abroadCompletionStatus, incomeCompletionStatus) = (getStatus(Abroad), getStatus(Income))
+    val (abroadCompletionStatus, incomeCompletionStatus, expensesCompletionStatus) = (
+      getStatus(Abroad),
```
lukastymo, line 48: getStatus return Option[Boolean] is hard to undersatnd by reading the method name and checking the return type what it express.

can we return proper Status like Completed or InProgress here?

app/viewmodels/TradeJourneyStatusesViewModel.scala
```diff
@@ -43,9 +44,13 @@ object TradeJourneyStatusesViewModel {
...
+      getStatus(Income),
+      getStatus(ExpensesTailoring)
+    )
 
-    val (abroadUrlString, incomeUrlString) = {
+    val (abroadUrlString, incomeUrlString, expensesUrlString) = {
```
lukastymo, line 53: this can be separate private method

app/viewmodels/TradeJourneyStatusesViewModel.scala
```diff
@@ -55,27 +60,38 @@ object TradeJourneyStatusesViewModel {
...
       (
         abroadUrl,
-        if (abroadCompletionStatus.getOrElse(false)) incomeUrl else "#"
+        if (abroadCompletionStatus.getOrElse(false)) incomeUrl else "#",
+        if (abroadCompletionStatus.getOrElse(false)) expensesUrl else "#"
       )
     }
 
-    val (abroadStatusString, incomeStatusString) =
+    val (abroadStatusString, incomeStatusString, expensesStatusString) =
```
lukastymo, line 75: separate private method.

let's not use .get, it's similar what I mentioned in .head

app/viewmodels/TradeJourneyStatusesViewModel.scala
```diff
@@ -43,9 +44,13 @@ object TradeJourneyStatusesViewModel {
       if (optJourney.isEmpty) None else optJourney.head.completedState
```
lukastymo, line 44: let's not use head, as list may be empty. even though you checked isEmpty, the code may change and the head will stay then. it's better to use map / haedOption etc.

app/viewmodels/TradeJourneyStatusesViewModel.scala
```diff
@@ -55,27 +60,38 @@ object TradeJourneyStatusesViewModel {
...
-    val (abroadStatusString, incomeStatusString) =
+    val (abroadStatusString, incomeStatusString, expensesStatusString) =
       (
         if (abroadCompletionStatus.isEmpty) notStartedStatus
         else if (abroadCompletionStatus.get) completedStatus
         else inProgressStatus,
         if (!abroadCompletionStatus.getOrElse(false)) cannotStartYetStatus
         else if (incomeCompletionStatus.isEmpty) notStartedStatus
         else if (incomeCompletionStatus.get) completedStatus
+        else inProgressStatus,
```
lukastymo, line 83: once you do the above, I think this block will be obvious how to write simpler

lukastymo (changes requested):

app/viewmodels/TradeJourneyStatusesViewModel.scala
Jake-Raffe: This will be done in a new PR making richer types in the backend and frontend.

lukastymo (approved):

lukastymo (approved):
Approving as it will unblock other journeys from being tested. Let's address my comment in separate PR

lukastymo (changes requested):
Let's try to address comments and merge it then. I think we can test sub-genre now by typing the URL address

lukastymo:
image
image

but when we use different imports Intellij is happy:
image

image

app/viewmodels/TradeJourneyStatusesViewModel.scala
```diff
@@ -36,56 +38,33 @@ object TradeJourneyStatusesViewModel {
   private val notStartedStatus     = "notStarted"
   private val cannotStartYetStatus = "cannotStartYet"
 
-  def buildSummaryList(business: TradesJourneyStatuses, taxYear: Int)(implicit messages: Messages): SummaryList = {
+  def buildSummaryList(tradesJourneyStatuses: TradesJourneyStatuses, taxYear: Int)(implicit messages: Messages): SummaryList = {
```
lukastymo, line 41: cyclomatic complexity is still too high

image

lukastymo (approved):

lukastymo (approved):

PR #92, opened 2023-11-21
[NOJIRA] Remove repeated messages

Jake-Raffe:


app/controllers/journeys/abroad/SelfEmploymentAbroadCYAController.scala
```diff
@@ -41,13 +41,12 @@ class SelfEmploymentAbroadCYAController @Inject() (override val messagesApi: Mes
...
   def onPageLoad(taxYear: Int, businessId: String): Action[AnyContent] = (identify andThen getData andThen requireData) { implicit request =>
-    val isAgent         = request.user.isAgent
-    val summaryListRows = SelfEmploymentAbroadSummary.row(taxYear, isAgent, businessId, request.userAnswers)
+    val summaryListRows = SelfEmploymentAbroadSummary.row(taxYear, request.user.userType, businessId, request.userAnswers)
     val summaryList     = SummaryList(Seq(summaryListRows))
 
     val nextRoute = nextPageUrl(taxYear, businessId, navigator)
 
-    Ok(view(taxYear, summaryList, nextRoute, isAgent))
+    Ok(view(taxYear, summaryList, nextRoute, request.user.isAgent))
```
lukastymo, line 49: We should move away from the boolean isAgent.

What would happen in terms of cost of a change if a 3rd type of users was introduced?: Individual, Agent, and... let's say Reviewer? 

You can use `user.userType` and then to String in your view, then you don't need any ifs

lukastymo (approved):

app/controllers/journeys/abroad/SelfEmploymentAbroadCYAController.scala
Jake-Raffe: Added to the tech debt

lukastymo (approved):

PR #102, opened 2023-11-23
[SASS-6311] Create StaffCostsDisallowableAmount page

Jake-Raffe:


app/navigation/ExpensesNavigator.scala
```diff
@@ -41,94 +42,108 @@ import javax.inject.{Inject, Singleton}
...
-              case _ =>
-                standard.routes.JourneyRecoveryController.onPageLoad()
-            }
+            _ =>
+              userAnswers.get(OfficeSuppliesPage, Some(businessId.value)) match {
+                case Some(OfficeSupplies.YesAllowable) =>
+                  officeSupplies.routes.OfficeSuppliesCYAController.onPageLoad(taxYear, businessId.value)
+                case Some(OfficeSupplies.YesDisallowable) =>
+                  officeSupplies.routes.OfficeSuppliesDisallowableAmountController.onPageLoad(taxYear, businessId.value, NormalMode)
+                case _ =>
```
lukastymo, line 57: it generally correct to use _, but if you have few cases it is safer to name them
```
None | Some(OfficeSupplies.No) =>
```

what you get is that in future when you have to add one more option, the place you forgot to handle will not compile and you will have to make concious decision what you should do for the new type.

lukastymo (dismissed):

PR #118, opened 2023-11-29
SASS-6355: FE implementation of save and continue for Income CYA - happy path

harry-ritchie:


app/utils/Logging.scala
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package utils
+
+import play.api.Logger
+
+trait Logging {
+  implicit lazy val logger: Logger = Logger(this.getClass)
```
lukastymo, line 22: oh nice one, lazy, so it will take the class name not the trait name I guess

PR #119, opened 2023-11-30
SASS-6366: Office Supplies CYA save and continue - happy path

harry-ritchie:


app/controllers/package.scala
```diff
@@ -31,4 +38,15 @@ package object controllers {
...
 
+  // Redirection to journey recovery on downstream error retrieval is a temporary action until we pick up the unhappy
+  // path tickets (JIRA TBA).
+  def submitJourneyAnswers[T](answers: T, context: SubmissionContext, service: SendJourneyAnswersService)(implicit
+      ec: ExecutionContext,
+      hc: HeaderCarrier,
+      writes: Writes[T]): Future[Result] =
+    (for {
+      _ <- EitherT(service.sendJourneyAnswers(context, answers))
+    } yield Redirect(SectionCompletedStateController.onPageLoad(context.taxYear, context.businessId, context.journey.toString, NormalMode)))
```
lukastymo, line 49: I think the common way of redirecting from any non error scenario is via Navigator. 

if you want a common helper method maybe you can pass onSuccess: Call (rendered by navigator nextCall=)

for leftMap can you reuse handleResult so we have one place where we log an error and redirect to recovery page?

lukastymo (approved):

app/controllers/package.scala
lukastymo: ok I move it back, having common place for sectioncompleted redirection may be better

lukastymo (approved):

lukastymo (approved):

PR #123, opened 2023-12-01
SASS-6363: Submit income answers

lukastymo:


app/models/common/JourneyContext.scala
```diff
@@ -23,8 +23,14 @@ sealed trait JourneyContext {
   val businessId: BusinessId
   val mtditid: Mtditid
   val journey: Journey
+
+  val answersUrl: String
 }
 
-case class SubmissionContext(taxYear: TaxYear, nino: Nino, businessId: BusinessId, mtditid: Mtditid, journey: Journey) extends JourneyContext
+case class JourneyAnswersWithNino(taxYear: TaxYear, nino: Nino, businessId: BusinessId, mtditid: Mtditid, journey: Journey) extends JourneyContext {
```
harry-ritchie, line 30: How about renaming to `JourneyContextWithNino`, and the case class below to `JourneyContext`?

harry-ritchie (approved):

PR #136, opened 2023-12-07
SASS-6481: Load prior data for Income journey

lukastymo:
It loads the submitted answers. The submitted answers are also called 'journey answers'. There is small logic for merging the submitted answers with the user session (called user answers).

The logic only is triggered on CYA page and only when there is no data for the journey in session.

The logic is:
- no session data + no submitted answers == no data
- no session data + submitted answers == submitted answers
- session data for journey A + submitted answers for journey B == answers merged A + B
- session data for journey A + submitted answers A == session data for journey A

Also few improvements for everyday development:
- I removed wartremover, as it caused more problems than helped
- You can now add 'PLAY_ENV=Dev' to your env /intellij env variables, refresh SBT and then you'll have no warn failure checks in intellij.


app/connectors/ContentHttpReads.scala
```diff
@@ -0,0 +1,27 @@
...
+package connectors
+
+import cats.implicits._
+import connectors.httpParser.JourneyStateParser.pagerDutyError
+import play.api.libs.json.Reads
+import uk.gov.hmrc.http.{HttpReads, HttpResponse}
+
+class ContentHttpReads[A: Reads] extends HttpReads[ContentResponse[Option[A]]] {
+  override def read(method: String, url: String, response: HttpResponse): ContentResponse[Option[A]] =
+    if (isSuccess(response.status)) response.json.validate[A].asOpt.asRight else Left(pagerDutyError(response))
```
harry-ritchie, line 26: Just thinking, could we simplify it and not return our right wrapped in an Option? 

```scala
class ContentHttpReads[A: Reads] extends HttpReads[ContentResponse[A]] {
  override def read(method: String, url: String, response: HttpResponse): ContentResponse[A] =
    if (isSuccess(response.status)) {
      response.json.validate[A].fold(_ => HttpError.`someParsingError`.asLeft, _.asRight)
    }
    else {
      Left(pagerDutyError(response))
    }
}

```

app/controllers/actions/SubmittedDataRetrievalActionImpl.scala
```diff
@@ -0,0 +1,93 @@
...
+        request.copy(userAnswers = if (updated.isEmpty) None else Some(updated))
+      }
+    }
+  }
+
+  private def upsertJourneyAnswers[A](ctx: JourneyContext, request: Request[A], existingAnswers: UserAnswers): Future[UserAnswers] = {
+    val result: ApiResultT[UserAnswers] = for {
+      maybeJourneyAnswers <- selfEmploymentService.getSubmittedAnswers[SubsetOfAnswers](ctx)(implicitly[Format[SubsetOfAnswers]], hc(request))
+      updatedAnswers <- maybeJourneyAnswers.fold {
+        EitherT.pure[Future, HttpError](existingAnswers)
```
harry-ritchie, line 67: Also, if are able to take out the Option like my comment above says, it will simplify this a little we won't need to accommodate for None case. What do you think? I may be missing something.

app/models/database/UserAnswers.scala
```diff
@@ -45,6 +46,26 @@ final case class UserAnswers(id: String, data: JsObject = Json.obj(), lastUpdate
...
+    val updatedAnswers =
+      data.transform(upsertFragment) match {
+        case JsSuccess(jsValue, _) => Success(jsValue)
+        case JsError(errors)       => Failure(JsResultException(errors))
+      }
+
+    updatedAnswers.map { updatedData =>
+      copy(data = updatedData, lastUpdated = Instant.now)
+    }
+  }
```
harry-ritchie, line 67: I may well me wrong here so do check, but is this code not equivalent to just doing the below:

```scala
    val existingAnswerData = (data \ businessId.value).asOpt[JsObject].getOrElse(JsObject.empty)
    val updatedData            = data + (businessId.value -> (existingAnswerData ++ dataFragment))

    copy(data = updatedData)
``` 

If so, it also gets rid of the Try.

app/models/requests/DataRequest.scala
```diff
@@ -33,6 +33,11 @@ case class OptionalDataRequest[A](request: Request[A], userId: String, user: Use
   val mtditid: Mtditid     = Mtditid(user.mtditid)
 }
 
+object OptionalDataRequest {
+  def empty(request: IdentifierRequest[_], user: User): OptionalDataRequest[_] =
```
harry-ritchie, line 37: This looks unused to me. Do we need it?

test/controllers/actions/SubmittedDataRetrievalActionImplSpec.scala
```diff
@@ -0,0 +1,114 @@
...
+import play.api.mvc.AnyContentAsEmpty
+import play.api.test.FakeRequest
+import stubs.repositories.StubSessionRepository
+import stubs.services.SelfEmploymentServiceStub
+import uk.gov.hmrc.auth.core._
+
+import java.time.Instant
+import scala.concurrent.ExecutionContext.Implicits.global
+
+class SubmittedDataRetrievalActionImplSpec extends AnyWordSpecLike with Matchers {
```
harry-ritchie, line 39: Would it be worth having some ITs for this class at some point? Could check what is in mongo and use wiremock for BE.

test/models/database/UserAnswersSpec.scala
```diff
@@ -0,0 +1,108 @@
...
+
+      result.data shouldBe Json.obj(
+        businessId.value -> Json.obj(
+          "key1" -> "value1",
+          "key2" -> "valueUpdated"
+        )
+      )
+    }
+
+    "merge two jsons if they have more than one business ids" in {
```
harry-ritchie, line 78: Maybe phrase as `merge json fragment with the correct businessId in the presence of multiple businessIds`

app/connectors/ContentHttpReads.scala
lukastymo: good point. Let's not mask the error in option but return left of error when it fails

app/controllers/actions/SubmittedDataRetrievalActionImpl.scala
lukastymo: None is when we don't have submitted values (request OK, but just no answers)

app/models/database/UserAnswers.scala
lukastymo: good one

app/models/requests/DataRequest.scala
lukastymo: removed

test/controllers/actions/SubmittedDataRetrievalActionImplSpec.scala
lukastymo: Not sure.

I don't think we need many ITs, they slow down refactoring. If the codebase is young and changing, they don't give much value. But once we have Income and Expenses entire journey we may start thinking if there is a chance to catch more bugs by adding more integration tests. At the end of the day we have Regression Tests: Journey UI which will catch a lot of bugs and protect us from breaking the code.

I would say once we have stable codebase + 100% unit test coverage + Regression Journey UI tests, then we can start thinking on integration tests. Or altenatively, if the logic between 2 components is tricky we may also add it earlier.

test/models/database/UserAnswersSpec.scala
lukastymo: too wordy, I like short and general title for tests - make refactoring easier.

`merge two jsons with the correct businessId if many businessIds present` ?

harry-ritchie (dismissed):

harry-ritchie (approved):

PR #151, opened 2023-12-15
[NOJIRA] Relevant Expenses journeys are available once Tailoring journey is Complete or InProgress

Jake-Raffe:


app/viewmodels/TradeJourneyStatusesViewModel.scala
```diff
@@ -106,34 +127,33 @@ object TradeJourneyStatusesViewModel {
...
-      journeyStatuses: TradesJourneyStatuses): JourneyStatus =
-    statusFromCompletedState(getCompletedState(journeyStatuses, journey)) match {
-      case NotStarted            => conditionalJourneyIsPassed(conditionalCompletedJourney)
-      case status: JourneyStatus => status
+  private def dependentJourneyIsInProgressOrCompleted(dependentJourney: Journey, checkOnlyCompleted: Boolean = false)(implicit
+      journeyStatuses: TradesJourneyStatuses): Boolean =
+    getJourneyStatus(dependentJourney) match {
+      case Completed | InProgress if !checkOnlyCompleted => true
+      case Completed                                     => true
+      case _                                             => false
```
lukastymo, line 135: it's good idea to not use `_` if it's a small number of elements.

Then if we add new sealed value, the code will not compile until we consciously change each of these pattern matches. Otherwise, when we add new value the code will compile which is not good, because may lead to some bugs.

lukastymo (approved):

lukastymo (approved):

PR #155, opened 2023-12-18
[SASS-6226] Limit BigDecimal form inputs to 2 decimal places

Jake-Raffe:


lukastymo:
Great PR, more lines removed than added!

app/forms/mappings/Formatters.scala
```diff
@@ -102,6 +102,33 @@ trait Formatters {
         baseFormatter.unbind(key, value.toString)
     }
 
+  private[mappings] def currencyFormatter(requiredKey: String, nonNumericKey: String, args: Seq[String] = Seq.empty): Formatter[BigDecimal] =
```
lukastymo, line 105: should we create FormattersSpec and test currencyFormatter in isolation. I'm interested in more edge cases scenarios. I think right now we have just general controller tests. But I'm thinking of more cases just for this function, it should be easy testable.

lukastymo (dismissed):

app/forms/mappings/Formatters.scala
Jake-Raffe: Pushed up now

lukastymo (approved):

conniebernardin (approved):

PR #157, opened 2023-12-19
[SASS-6064] Create Advertising and Marketing pages

Jake-Raffe:


app/controllers/journeys/expenses/advertisingAndMarketing/AdvertisingAmountController.scala
```diff
@@ -0,0 +1,81 @@
...
+      }
+    Ok(view(preparedForm, mode, user, taxYear, businessId))
+  }
+
+  def onSubmit(taxYear: TaxYear, businessId: BusinessId, mode: Mode): Action[AnyContent] = (identify andThen getData) async { implicit request =>
+    selfEmploymentService.getAccountingType(request.user.nino, businessId, request.user.mtditid) flatMap {
+      case Left(_) => Future.successful(Redirect(JourneyRecoveryController.onPageLoad()))
+      case Right(accountingType) =>
+        val user = request.userType
+        formProvider(user)
```
lukastymo, line 65: can we make it a bit less nested. You can split it into smaller method(s) so we don't have for inside fold.

if we need to nest something to higher levels it usually means Another Responsibility. our method ideally should do one thing.

If we start splitting large methods into smaller we may observe reusable patterns.

app/viewmodels/checkAnswers/expenses/advertisingAndMarketing/AdvertisingDisallowableAmountSummary.scala
```diff
@@ -0,0 +1,69 @@
...
+          "site.change",
+          expenses.advertisingAndMarketing.routes.AdvertisingDisallowableAmountController.onPageLoad(taxYear, businessId, CheckMode).url)
+          .withVisuallyHiddenText(messages("advertisingDisallowableAmount.change.hidden"))
+      )
+    )
+
+  private def areAnyAdvertisingDisallowable(advertisingAnswer: AdvertisingOrMarketing): Boolean =
+    advertisingAnswer match {
+      case YesDisallowable => true
+      case _               => false
```
lukastymo, line 66: case YesAllowable | No => false

lukastymo (approved):

PR #174, opened 2023-12-29
[NoJira] (1/2) Remove session repository from Controllers

Jake-Raffe:


harry-ritchie (approved):

app/controllers/journeys/expenses/tailoring/individualCategories/DisallowableInterestController.scala
```diff
@@ -24,15 +24,15 @@ import navigation.ExpensesTailoringNavigator
...
 import play.api.mvc.{Action, AnyContent, MessagesControllerComponents}
-import repositories.SessionRepository
+import services.SelfEmploymentService
 import uk.gov.hmrc.play.bootstrap.frontend.controller.FrontendBaseController
 import views.html.journeys.expenses.tailoring.individualCategories.DisallowableInterestView
 
 import javax.inject.Inject
 import scala.concurrent.{ExecutionContext, Future}
 
 class DisallowableInterestController @Inject() (override val messagesApi: MessagesApi,
```
harry-ritchie, line 34: Looks like in the UTs we are sometimes mocking the session repository, and other times the service. Maybe we should be consistent? Maybe not something we need to do in this PR, but we'll probs want to use the base classes at some point to reduce all the duplication so maybe we can sort that out then?

PR #193, opened 2024-01-08
[NoJira] (2/3) Remove session repository from Controllers

Jake-Raffe:


app/controllers/journeys/expenses/goodsToSellOrUse/DisallowableGoodsToSellOrUseAmountController.scala
```diff
@@ -35,7 +35,7 @@ import scala.concurrent.{ExecutionContext, Future}
 
 @Singleton
 class DisallowableGoodsToSellOrUseAmountController @Inject() (override val messagesApi: MessagesApi,
-                                                              sessionRepository: SessionRepository,
+                                                              selfEmploymentService: SelfEmploymentService,
```
harry-ritchie, line 38: It may be better to inject the abstraction rather than the impl itself, as it gives us more flexibility to swap out/test.

app/controllers/journeys/expenses/tailoring/individualCategories/FinancialExpensesController.scala
```diff
@@ -79,21 +83,37 @@ class FinancialExpensesController @Inject() (override val messagesApi: MessagesA
...
-              formWithErrors => Future.successful(BadRequest(view(formWithErrors, mode, request.userType, taxYear, businessId, accountingType))),
-              value =>
-                for {
-                  clearedAnswers <- Future.fromTry(clearPageDataFromUserAnswers(request.userAnswers, Some(businessId), value))
-                  updatedAnswers <- Future.fromTry(clearedAnswers.set(FinancialExpensesPage, value, Some(businessId)))
-                  _              <- sessionRepository.set(updatedAnswers)
-                } yield Redirect(navigator.nextPage(FinancialExpensesPage, mode, updatedAnswers, taxYear, businessId))
-            )
-      }
+      def handleError(formWithErrors: Form[_], userType: UserType, accountingType: AccountingType): Future[Result] =
```
harry-ritchie, line 86: We could just inline this on line 104, it is a one liner and splitting up doesn't make things any clearer imo.

app/controllers/journeys/expenses/tailoring/individualCategories/FinancialExpensesController.scala
```diff
@@ -79,21 +83,37 @@ class FinancialExpensesController @Inject() (override val messagesApi: MessagesA
...
+        Future.successful(
+          BadRequest(view(formWithErrors, mode, userType, taxYear, businessId, accountingType))
+        )
+
+      def handleSuccess(userAnswers: UserAnswers, answers: Set[FinancialExpenses]): Future[Result] =
```
harry-ritchie, line 91: Could we maybe rename answers somehow to something a bit more distinct e.g. `financialExpenses` or wtte? Lots of ..answers here, `clearedAnswers`, `answers`, `userAnswers` etc.

app/controllers/journeys/expenses/goodsToSellOrUse/DisallowableGoodsToSellOrUseAmountController.scala
Jake-Raffe: Happy for me to do this in a new PR where I switch it everywhere to the BaseService?

test/controllers/journeys/expenses/goodsToSellOrUse/DisallowableGoodsToSellOrUseAmountControllerSpec.scala
```diff
@@ -58,14 +54,7 @@ class DisallowableGoodsToSellOrUseAmountControllerSpec
...
       .value
 
-  private val mockSelfEmploymentService = mock[SelfEmploymentService]
-
-  override val bindings: List[Binding[_]] = List(
-    bind[ExpensesNavigator].toInstance(new FakeExpensesNavigator(onwardRoute)),
-    bind[SelfEmploymentService].toInstance(mockSelfEmploymentService)
-  )
-
-  when(mockSelfEmploymentService.persistAnswer(anyBusinessId, anyUserAnswers, any, any)(any)) thenReturn Future.successful(filledUserAnswers)
```
harry-ritchie, line 68: Removing this works for now, as in the `ControllerSpec` we are mocking the session repository. I think ideally we would mock the service instead, so when the session repository mock gets removed in your TODO in `ControllerSpec`, we may need to replace it with a service mock.

app/controllers/journeys/expenses/goodsToSellOrUse/DisallowableGoodsToSellOrUseAmountController.scala
harry-ritchie: Yep no probs

harry-ritchie (approved):

PR #196, opened 2024-01-09
NOJIRA: Use common code across all Income journey controllers

harry-ritchie:
PR to use the common package object code across the income controllers, as is done throughout most of the expenses controllers. As a result, controllers do not interact directly with the repository layers, better separating concerns.


Jake-Raffe (approved):
Much better

PR #200, opened 2024-01-09
[NOJIRA] Use SelfEmploymentService base trait instead of implementation class

Jake-Raffe:


app/services/SelfEmploymentService.scala
```diff
@@ -86,11 +83,10 @@ object SelfEmploymentService {
 
   private val maxAllowance = BigDecimal(1000.00)
 
-  // TODO: Return an error as left once we have impl. error handling instead of redirecting.
-  def getMaxTradingAllowance(businessId: BusinessId, userAnswers: UserAnswers): Either[Result, BigDecimal] =
+  def getMaxTradingAllowance(businessId: BusinessId, userAnswers: UserAnswers): Either[NotFoundError, BigDecimal] =
     userAnswers
       .get(TurnoverIncomeAmountPage, Some(businessId))
-      .fold(redirectJourneyRecovery().asLeft[BigDecimal]) { turnover =>
+      .fold(NotFoundError("TurnoverIncomeAmount not found").asLeft[BigDecimal]) { turnover =>
```
harry-ritchie, line 89: I'm not sure this `NotFoundError` is meant for this type of case where we cannot find data for a page. I think maybe leave it as it was before, as we have not implemented error handling properly yet.

harry-ritchie (approved):

PR #247, opened 2024-02-09
SASS 6997: Create Workplace Running Costs CYA

conniebernardin:


app/controllers/journeys/expenses/workplaceRunningCosts/WorkplaceRunningCostsCYAController.scala
```diff
@@ -0,0 +1,133 @@
...
+    val months51To100 = request.getValue(WorkingFromHomeHours51To100, businessId)
+    val months101Plus = request.getValue(WorkingFromHomeHours101Plus, businessId)
+
+    val months1Person = request.getValue(LivingAtBusinessPremisesOnePerson, businessId)
+    val months2People = request.getValue(LivingAtBusinessPremisesTwoPeople, businessId)
+    val months3People = request.getValue(LivingAtBusinessPremisesThreePlusPeople, businessId)
+
+    (months1Person, months2People, months3People, months25To50, months51To100, months101Plus) match {
+      case (Some(months1Person), Some(months2People), Some(months3People), Some(months25To50), Some(months51To100), Some(months101Plus)) =>
+        val amount25To50  = months25To50 * 10
```
lukastymo, line 118: I think it would be better to extract these calculations to some case class / object, so it can be independently unit tested.  It is not immediately obvious what these are.

lukastymo (approved):

PR #259, opened 2024-02-21
[SASS-7369] Create ZeroEmissionGoodsVehicle page

Jake-Raffe:


app/forms/capitalallowances/zeroEmissionGoodsVehicle/ZeroEmissionGoodsVehicleFormProvider.scala
```diff
@@ -0,0 +1,32 @@
...
+import models.common.{TaxYear, UserType}
+import play.api.data.Form
+
+import javax.inject.Inject
+
+class ZeroEmissionGoodsVehicleFormProvider @Inject() extends Mappings {
+
+  def apply(userType: UserType, taxYear: TaxYear): Form[Boolean] =
+    Form(
+      "value" -> boolean(s"zeroEmissionGoodsVehicle.error.required.$userType", args = Seq((taxYear.value - 1).toString, taxYear.value.toString))
```
lukastymo, line 29: small improvement suggestion: it's not immediately obvious what is `taxYear.value -1` and we use it quite often.

We can add a helper method inside TaxYear somethlig like:
```
final case class TaxYear(value: int) extends AnyVal {
  val startOfTheYear: Int = value - 1
}
```

app/controllers/journeys/capitalallowances/zeroEmissionGoodsVehicle/ZeroEmissionGoodsVehicleController.scala
```diff
@@ -0,0 +1,93 @@
...
+              updatedUserAnswers                <- service.persistAnswer(businessId, editedUserAnswers, answer, ZeroEmissionGoodsVehiclePage)
+            } yield Redirect(navigator.nextPage(ZeroEmissionGoodsVehiclePage, redirectMode, updatedUserAnswers, taxYear, businessId))
+        )
+  }
+
+  private def handleGatewayQuestion(currentAnswer: Boolean,
+                                    request: DataRequest[_],
+                                    mode: Mode,
+                                    businessId: BusinessId): Future[(UserAnswers, Mode)] = {
+    val pagesToBeCleared: List[Settable[_]] = List()
```
lukastymo, line 81: More idiomatic way to create empty list is `= Nil`

lukastymo (approved):

app/forms/capitalallowances/zeroEmissionGoodsVehicle/ZeroEmissionGoodsVehicleFormProvider.scala
Jake-Raffe: Good idea, will do

app/controllers/journeys/capitalallowances/zeroEmissionGoodsVehicle/ZeroEmissionGoodsVehicleController.scala
Jake-Raffe: Leaving an empty list to remember to add each page into the list as each page is created

lukastymo: ok

lukastymo (dismissed):

lukastymo (approved):

PR #272, opened 2024-02-27
[NOJIRA] Create Controller package spec

Jake-Raffe:


test/controllers/ControllersPackageSpec.scala
```diff
@@ -0,0 +1,59 @@
...
+        val commencementYear = 2018
+        val result           = getMaxMonthsWithinTaxYearOrRedirect(aBusinessData.copy(commencementDate = Some(s"$commencementYear-04-06")), taxYear)
+        result shouldBe defaultMaxMonths.asRight
+      }
+      "should return the number of months between the Commencement date and TaxYear.endYear as Right" in {
+        val taxYearCutoffDate = LocalDate.parse(s"${taxYear.endYear}-04-05")
+        val commencementDate  = s"${taxYear.endYear}-01-05"
+        val expectedResult    = ChronoUnit.MONTHS.between(LocalDate.parse(commencementDate), taxYearCutoffDate).toInt
+        val result            = getMaxMonthsWithinTaxYearOrRedirect(aBusinessData.copy(commencementDate = Some(commencementDate)), taxYear)
+        result shouldBe expectedResult.asRight
```
lukastymo, line 43: just use hardcode values. Tests should be as simple as possible, we already have same line in the code under test: ChronoUnit.MONTHS.... 

it's perfectly fine to just use result shouldBe 3.asRight

lukastymo (approved):

test/controllers/ControllersPackageSpec.scala
Jake-Raffe: good to know, I'll remove it in my current PR, thanks

PR #283, opened 2024-03-01
SAAS-7420 Capital Allowances - Goods Vehicle journey

lukastymo:
I did few refactorings:
- disabled checking duplicate of values in messages.en - it is too annoying, and time consuming. We can always do it at the end
- added test which check if we don't add duplicated properties - I saw couple of them, that's dangerous as in the Map only one is taken and rest is ignored
- added view component for the conditioanl radio so we don't have to use javascript
- I moved away from Navigator, as a) one big navigator cause a lot of conflicts when we work on different journeys b) it is much better to just decide when to go on the Page level in Page case class
- replaced Yes, No enum to true false


app/models/journeys/Journey.scala
```diff
@@ -157,12 +157,12 @@ object Journey extends Enum[Journey] with utils.PlayJsonEnum[Journey] {
...
-      ZecUseOutsideSEPage.pageName,
-      ZecUseOutsideSEPercentagePage.pageName,
-      ZecHowMuchDoYouWantToClaimPage.pageName,
-      ZecClaimAmount.pageName
+      ZegvTotalCostOfVehiclePage.pageName,
+      ZegvOnlyForSelfEmploymentPage.pageName,
+      ZegvUseOutsideSEPage.pageName,
+      ZegvUseOutsideSEPercentagePage.pageName,
+      ZegvHowMuchDoYouWantToClaimPage.pageName,
+      ZegvClaimAmountPage.pageName
```
Jake-Raffe, line 165: These should be in GoodsVehicle, not cars. Strangely, all the ones for vehicles have disappeared outside of this PR

app/pages/capitalallowances/zeroEmissionGoodsVehicle/ZegvOnlyForSelfEmploymentPage.scala
```diff
@@ -0,0 +1,47 @@
...
+import controllers.standard
+import pages.redirectOnBoolean
+
+object ZegvOnlyForSelfEmploymentPage extends ZegvBasePage[Boolean] {
+  override def toString: String = "zegvOnlyForSelfEmployment"
+
+  override val dependentPagesWhenYes: List[Settable[_]] =
+    List(
+      ZegvUseOutsideSEPercentagePage
+    )
```
Jake-Raffe, line 34: Needs to clear the ZegvUseOutsideSEPage as well

Jake-Raffe (dismissed):

app/pages/capitalallowances/zeroEmissionGoodsVehicle/ZegvOnlyForSelfEmploymentPage.scala
lukastymo: done

harry-ritchie (approved):

PR #322, opened 2024-03-21
SASS-5912 page hopping for expense tailroing

lukastymo:


app/models/journeys/Journey.scala
```diff
@@ -100,6 +109,10 @@ object Journey extends Enum[Journey] with utils.PlayJsonEnum[Journey] {
     override val pageKeys: List[PageName] = List(
       ExpensesCategoriesPage.pageName
     )
+    override def startUrl(taxYear: TaxYear, businessId: BusinessId, mode: Mode): String =
+      expenses.tailoring.routes.ExpensesCategoriesController.onPageLoad(taxYear, businessId, mode).url
+
+    override def startPage: QuestionPage[_] = ExpensesCategoriesPage
   }
```
Jake-Raffe, line 116: we'll need to come up with a solution for the WorkplaceRunningCosts journey were there are two potential start pages

lukastymo: good point. I will have a look

lukastymo: will get back to it later, but I allow now to pass start page in the HopChecker

Jake-Raffe (approved):

PR #335, opened 2024-03-27
[SASS-7255] Fix various journey issues

Jake-Raffe:


app/forms/mappings/Formatters.scala
```diff
@@ -25,14 +25,21 @@ import scala.util.control.Exception.nonFatalCatch
...
       override def bind(key: String, data: Map[String, String]): Either[Seq[FormError], String] =
         data.get(key) match {
           case None                      => Left(Seq(FormError(key, errorKey, args)))
           case Some(s) if s.trim.isEmpty => Left(Seq(FormError(key, errorKey, args)))
-          case Some(s)                   => Right(if (toUpperCase) s.toUpperCase else s)
+          case Some(s) =>
+            Right {
+              val stripped = if (stripWhitespace) s.trim().replaceAll("\\s", "") else s
+              if (toUpperCase) stripped.toUpperCase else stripped
+            }
```
Jake-Raffe, line 42: Is there a better way to do this?

lukastymo: matter of taste, but maybe something like this:

```
import scala.util.chaining._
import cats.implicits._

val normalized =
  s
    .pipe(str => if (stripWhitespace) str.trim.replaceAll("\\s+", "") else str)
    .pipe(str => if (toUpperCase) str.toUpperCase else str)

normalized.asRight
```

lukastymo (approved):

app/forms/mappings/Formatters.scala
Jake-Raffe: thanks, I'll add this in another PR

PR #353, opened 2024-04-09
[SASS-6247] Upgrade Expenses journeys - 2

Jake-Raffe:


test/services/SelfEmploymentServiceSpec.scala
```diff
@@ -178,39 +188,94 @@ class SelfEmploymentServiceSpec extends SpecBase with MockitoSugar with Argument
...
+            taxYear,
+            NormalMode)
+          .futureValue
+
+      val expectedAnswers = existingZegvAnswers.set(ZeroEmissionGoodsVehiclePage, false, businessId.some).success.value
+      val dbAnswers       = repository.state(userAnswersId)
       assert(dbAnswers === expectedAnswers)
+
+      assert(result.header.status === SEE_OTHER)
+      assert(result.header.headers.head._2 === ZeroEmissionGoodsVehiclePage.cyaPage(taxYear, businessId).url)
```
harry-ritchie, line 251: We could use Play's built in helpers for this, where we pass in in the future:
```scala
status(result) shouldBe ...
redirectLocation(result) shouldBe ...
```

app/controllers/journeys/expenses/goodsToSellOrUse/DisallowableGoodsToSellOrUseAmountController.scala
```diff
@@ -17,67 +17,59 @@
...
+  private val page = DisallowableGoodsToSellOrUseAmountPage
 
-          Future.successful(Ok(view(preparedForm, mode, request.userType, taxYear, businessId, formatMoney(goodsAmount))))
-      }
+  def onPageLoad(taxYear: TaxYear, businessId: BusinessId, mode: Mode): Action[AnyContent] = (identify andThen getData andThen requireData) {
+    implicit request =>
+      (for {
+        allowableAmount <- request.valueOrRedirectDefault(GoodsToSellOrUseAmountPage, businessId)
+        form = fillForm(page, businessId, formProvider(request.userType, allowableAmount))
+      } yield Ok(view(form, mode, request.userType, taxYear, businessId, formatMoney(allowableAmount)))).merge
```
harry-ritchie, line 55: Not wrong by any means, but worth considering just explicitly using map if we are not sequencing any monadic operations.

test/controllers/actions/SubmittedDataRetrievalActionProviderImplSpec.scala
```diff
@@ -55,7 +56,7 @@ class SubmittedDataRetrievalActionProviderImplSpec extends AnyWordSpecLike with
 
       val underTest = new SubmittedDataRetrievalActionProviderImpl(connector, repo)
 
-      val result = underTest.loadTaskList(taxYear, fakeOptionalRequest).value.futureValue
```
harry-ritchie, line 58: Just out-of-interest, did you replace `futureValue` with `await` as it gave you a future not completed test failure?

test/services/SelfEmploymentServiceSpec.scala
Jake-Raffe: agreed, thanks

test/controllers/actions/SubmittedDataRetrievalActionProviderImplSpec.scala
Jake-Raffe: Yes, the test would start failing half the time from a timeout

app/controllers/journeys/expenses/goodsToSellOrUse/DisallowableGoodsToSellOrUseAmountController.scala
Jake-Raffe: Yep, think you're right. Will change them

harry-ritchie (approved):

PR #377, opened 2024-07-08
SASS-8995: Class 4 - Create empty pages 1-3

Dweik97:


conf/prod.routes
```diff
@@ -3,5 +3,5 @@
 ->        /update-and-submit-income-tax-return/self-employment        appAdjustments.Routes
 ->        /update-and-submit-income-tax-return/self-employment        appExpenses.Routes
 ->        /update-and-submit-income-tax-return/self-employment        appCapitalAllowances.Routes
-->        /update-and-submit-income-tax-return/self-employment        appAdjustments.Routes
+->        /update-and-submit-income-tax-return/self-employment        appNICs.Routes
```
Dweik97, line 6: L6 was a duplicate; appAdjustments.Routes already exists in L3.

app/controllers/journeys/nics/Class4ExemptionCategoryController.scala
```diff
@@ -0,0 +1,61 @@
...
+                                                   requireData: DataRequiredAction,
+                                                   formProvider: BooleanFormProvider,
+                                                   service: SelfEmploymentService,
+                                                   view: Class4ExemptionCategoryView)
+    extends FrontendBaseController
+    with I18nSupport {
+
+  private val page = Class4ExemptionCategoryPage
+
+  def onPageLoad(taxYear: TaxYear, businessId: BusinessId, mode: Mode): Action[AnyContent] = (identify andThen getData andThen requireData) {
```
lukastymo, line 48: is this National Insurance Contributions? If yes we don't want to pass businessId because it's common

lukastymo (changes requested):
image

it looks like this NIN journey is common, so we don't want to pass business id

app/pages/nics/Class4ExemptionCategoryPage.scala
```diff
@@ -0,0 +1,33 @@
...
+import models.database.UserAnswers
+import pages.OneQuestionPage
+import play.api.mvc.Call
+
+case object Class4ExemptionCategoryPage extends OneQuestionPage[Boolean] {
+  override def toString: String = "class4ExemptionCategory"
+
+  override def cyaPage(taxYear: TaxYear, businessId: BusinessId): Call = ???
+
+  override def nextPageInNormalMode(userAnswers: UserAnswers, businessId: BusinessId, taxYear: TaxYear): Call = cyaPage(taxYear, businessId)
```
Jake-Raffe, line 29: I'd link each to the next page so it has the default of running through the journey when you click continue. You can then add TODOs to add the logic for when it should navigate to one page or another. 
Can use the 'redirectOnBoolean' method to determine the next page for yes/no gateway pages.

app/pages/nics/Class4ExemptionCategoryPage.scala
```diff
@@ -0,0 +1,33 @@
...
+
+  override def hasAllFurtherAnswers(businessId: BusinessId, userAnswers: UserAnswers): Boolean =
+    userAnswers.get(this, businessId).isDefined
```
Jake-Raffe, line 32: Similar to above, would specify that hasAllFurtherAnswers is dependent on this page being answered and the next page's hasAllFurtherAnswers = true (or you can leave a todo with the ticket number of the content ticket for this page.

Can also override val dependentPagesWhenNo and list the two pages afterthis that are skipped if you answer 'No'. That will ensure that the data for those pages is deleted if the user comes back and answers No.

app/views/journeys/nics/Class4ExemptionCategoryView.scala.html
```diff
@@ -0,0 +1,46 @@
...
+    @if(form.errors.nonEmpty) {
+            @govukErrorSummary(ErrorSummaryViewModel(form))
+    }
+
+    @heading(messages(s"class4ExemptionCategory.title.$userType"), None, extraClasses = "no-padding-top govuk-!-margin-bottom-3")
+
+    @formHelper(action = Class4ExemptionCategoryController.onSubmit(taxYear, businessId, mode), 'autoComplete -> "off") {
+
+        @govukButton(ButtonViewModel(messages("site.continue")))
+
```
Jake-Raffe, line 44: There are some templates that are a bit cleaner we can use. I'd have a look at the profit and loss journey to see how the views and page objects are set up there

app/controllers/journeys/nics/Class4NICsController.scala
```diff
@@ -0,0 +1,61 @@
...
+      val form = fillForm(page, businessId, formProvider(page, request.userType))
+      Ok(view(form, taxYear, businessId, request.userType, mode))
+  }
+
+  def onSubmit(taxYear: TaxYear, businessId: BusinessId, mode: Mode): Action[AnyContent] = (identify andThen getData andThen requireData).async {
+    implicit request =>
+      def handleError(formWithErrors: Form[_]): Result = BadRequest(view(formWithErrors, taxYear, businessId, request.userType, mode))
+
+      service.defaultHandleForm(formProvider(page, request.userType), page, businessId, taxYear, mode, handleError)
+  }
```
Jake-Raffe, line 59: Have a look at GoodsAndServicesForYourOwnUseController submit method, which handles the boolean submission for a gateway question (will then clear data for skipped pages)

app/views/journeys/nics/Class4ExemptionCategoryView.scala.html
Dweik97: I wasn't aware we had all those clean components to choose from. I followed the design of a page with similar structure but looks like that was outdated.
Screenshot 2024-07-08 at 21 44 54

Jake-Raffe: We haven't had a chance yet to go back and make a lot of the older pages consistent with the new ones. It's a fairly minor point but worth changing if you're making changes anyway.
I'll probably make a few quick PRs this sprint to add some of these consistencies.

Jake-Raffe (approved):

PR #382, opened 2024-07-10
[SASS-8621] P&L - Create GoodsAndServicesAmount content page

Jake-Raffe:


app/views/components/HeadingWithCaption.scala.html
Dweik97, line 1: I feel rather than another Heading file, we should merge all into one an customise on use.

image

Dweik97 (approved):

app/views/components/HeadingWithCaption.scala.html
Jake-Raffe: Merging all will be a large undertaking because there's an almost even split of usage between the 3 of them. The new one will be what I use as a base for the combination of all. I shall have to do this in  a different ticket though

PR #393, opened 2024-07-15
SASS-8710: Content page - Class 4 NICs first page

Dweik97:


app/models/journeys/nics/NICsThresholds.scala
```diff
@@ -0,0 +1,71 @@
...
+  private case class Class4Limits(lowerProfitsLimit: Int, upperProfitsLimit: Int, rateBetweenLimits: Double, rateAboveUpperLimit: Double)
+
+  private def formatter(number: AnyVal): String = NumberFormat.getNumberInstance.format(number)
+
+  private def getFiguresForYear[T](taxYear: Int, figuresMap: Map[Int, T]): T = figuresMap.getOrElse(taxYear, figuresMap(fallbackYear))
+
+  object Class4NICsFigures {
+
+    private val taxYearFiguresMap: Map[Int, Class4Limits] = Map(
+      2022 -> Class4Limits(9568, 50270, 9, 2),
```
lukastymo, line 38: we may need to be able to go -4 years so 2021 may be also needed.

Also you can add TODO as we need centralized place for per-year config

And we may throw an exception (the app should not start) if we have missing tax year config, just to not fail in runtime

lukastymo (approved):

PR #403, opened 2024-07-22
[SASS-9625] Create CheckNetProfitLossPage content

Jake-Raffe:


app/models/journeys/adjustments/ProfitOrLoss.scala
```diff
@@ -0,0 +1,34 @@
...
+package models.journeys.adjustments
+
+import enumeratum._
+import models.common.{Enumerable, WithName}
+
+sealed trait ProfitOrLoss extends EnumEntry {
+  override def entryName: String = toString.toLowerCase
+}
+
+object ProfitOrLoss extends Enumerable.Implicits {
```
lukastymo, line 26: There is a more idiomatic (enumeratum way). Check how JourneyStatus is done:

object JourneyStatus extends Enum[JourneyStatus] with utils.PlayJsonEnum[JourneyStatus]

then you can use `  val values = findValues` and you have writes/reads for free

app/viewmodels/journeys/adjustments/NetBusinessProfitOrLossSummary.scala
```diff
@@ -0,0 +1,82 @@
...
+
+import models.journeys.adjustments.ProfitOrLoss
+import play.api.i18n.Messages
+import uk.gov.hmrc.govukfrontend.views.Aliases.Table
+import uk.gov.hmrc.govukfrontend.views.viewmodels.table.TableRow
+import viewmodels.checkAnswers.{buildTable, buildTableAmountRow}
+
+object NetBusinessProfitOrLossSummary {
+
+  private val placeholderAmount: BigDecimal = 200
```
lukastymo, line 27: does placeholder means it's just  for test because we don't have a backend? I would add clearly in the comment in next PR there will come backend and we will remove this value. 

Sometimes placeholder means a text which stays and is rendered for the user inputs. 

In general, I would loud and clear comment every single test data I commit to the production code

app/viewmodels/journeys/adjustments/NetBusinessProfitOrLossSummary.scala
```diff
@@ -0,0 +1,82 @@
...
+    val netProfitOrLoss            = turnover + incomeNotCountedAsTurnover + totalExpenses
+
+    val netProfitRows: Seq[Seq[TableRow]] = Seq(
+      buildTableAmountRow("profitOfLoss.turnover", turnover),
+      buildTableAmountRow("incomeNotCountedAsTurnover.title", incomeNotCountedAsTurnover),
+      buildTableAmountRow("profitOfLoss.totalExpenses", totalExpenses),
+      buildTableAmountRow(s"profitOfLoss.netProfitOrLoss.$profitOrLoss", netProfitOrLoss)
+    )
+
+    buildTable(
```
lukastymo, line 43: SRP: Single Responsible Principle - "every class, module, or function in a program should have one responsibility/purpose in a program."

here we're mixing calculations with the UI Table.

I suggest extracting  calculation to a separate function returning BigDecimal, which will be easier to refactor and test later.

app/viewmodels/journeys/adjustments/NetBusinessProfitOrLossSummary.scala
```diff
@@ -0,0 +1,82 @@
...
+      caption = Some(messages(s"profitOfLoss.netProfitOrLoss.$profitOrLoss")),
+      "govuk-!-margin-top-6 govuk-!-margin-bottom-9")
+  }
+
+  def buildAdditionsTable(profitOrLoss: ProfitOrLoss)(implicit messages: Messages): Table = {
+
+    val balancingCharge      = placeholderAmount
+    val goodsAndServices     = placeholderAmount
+    val disallowableExpenses = placeholderAmount
+    val totalAdditions       = balancingCharge + goodsAndServices + disallowableExpenses
```
lukastymo, line 55: same as above

app/viewmodels/journeys/adjustments/NetBusinessProfitOrLossSummary.scala
```diff
@@ -0,0 +1,82 @@
...
+    )
+
+    buildTable(None, additionsRows, caption = Some(messages(s"profitOfLoss.additions.$profitOrLoss")), "govuk-!-margin-bottom-9")
+  }
+
+  def buildDeductionsTable(profitOrLoss: ProfitOrLoss)(implicit messages: Messages): Table = {
+
+    val capitalAllowances  = placeholderAmount
+    val turnoverNotTaxable = placeholderAmount
+    val totalDeductions    = capitalAllowances + turnoverNotTaxable
```
lukastymo, line 71: and here

lukastymo (dismissed):

app/viewmodels/journeys/adjustments/NetBusinessProfitOrLossSummary.scala
Jake-Raffe: Have added a comment for whoever needs to do this in their ticket

app/assets/stylesheets/application.scss
```diff
@@ -1,7 +1,3 @@
-strong {
```
Jake-Raffe, line 1: This used to be needed for the tasklist but is no longer needed so best to remove, plus it's throwing off the formatting for this page

Jake-Raffe: Yes, it was added to all strong tags in the service. I have checked all pages with 'strong' and they all look as they should. Anthony has ACd

PR #404, opened 2024-07-22
SASS-8620 P&L: Goods and Services for own use

kuldeepwadhwa9:


app/controllers/journeys/adjustments/profitOrLoss/GoodsAndServicesForYourOwnUseController.scala
```diff
@@ -50,13 +51,17 @@ class GoodsAndServicesForYourOwnUseController @Inject() (override val messagesAp
 
   def onPageLoad(taxYear: TaxYear, businessId: BusinessId, mode: Mode): Action[AnyContent] = (identify andThen getData andThen requireData) {
     implicit request =>
-      val filledForm = fillForm(page, businessId, formProvider(page, request.userType))
-      Ok(view(filledForm, taxYear, businessId, request.userType, mode))
+      val accountingType = returnAccountingType(businessId)
```
Jake-Raffe, line 54: This probably doesn't need to be assigned to a val since it's very short and self explanatory it can just go straight in the view on line 56 (though this is a personal preference thing so the choice is yours!)

kuldeepwadhwa9: Done :)

lukastymo (dismissed):

lukastymo (approved):

Jake-Raffe (approved):

PR #417, opened 2024-08-07
SASS-8635 P&L:Loss brought forward - create taxYearBroughtForwardFrom…

kuldeepwadhwa9:
… content page 3

### DIFF
--- a/app/models/journeys/adjustments/WhichYearIsLossReported.scala
+++ b/app/models/journeys/adjustments/WhichYearIsLossReported.scala
@@ -27,10 +27,16 @@ object WhichYearIsLossReported extends Enumerable.Implicits {
 
   case object Year2018to2019 extends WithName("2018to2019") with WhichYearIsLossReported
   case object Year2019to2020 extends WithName("2019to2020") with WhichYearIsLossReported
+  case object Year2020to2021 extends WithName("2020to2021") with WhichYearIsLossReported
+  case object Year2021to2022 extends WithName("2021to2022") with WhichYearIsLossReported
+  case object Year2022to2023 extends WithName("2022to2023") with WhichYearIsLossReported
 
   val values: Seq[WhichYearIsLossReported] = Seq(
     Year2018to2019,
-    Year2019to2020
+    Year2019to2020,
+    Year2020to2021,
+    Year2021to2022,
+    Year2022to2023
   )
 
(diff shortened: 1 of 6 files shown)


Dweik97 (approved):

app/models/journeys/adjustments/WhichYearIsLossReported.scala
```diff
@@ -21,16 +21,24 @@ import play.api.i18n.Messages
...
 
+import java.time.Year
+
 sealed trait WhichYearIsLossReported
 
 object WhichYearIsLossReported extends Enumerable.Implicits {
 
-  case object Year2018to2019 extends WithName("2018to2019") with WhichYearIsLossReported
-  case object Year2019to2020 extends WithName("2019to2020") with WhichYearIsLossReported
+  case object currentTaxYearMinusFour  extends WithName(generateYears().last) with WhichYearIsLossReported
```
lukastymo, line 30: we changed hardcoded 2018to2019 with generateYears() - but it is still hardcoded. The application need to be restarted if we want to support more years. I wouldn't use generateYears as enum.

We have two choices:
- We hardcode like it was but for more years, and each year we remove the oldest one and add newer one. Not perfect, but easy to implement.
- We move away from enum as the value is dynamic always 5 tax years.

app/models/journeys/adjustments/WhichYearIsLossReported.scala
```diff
@@ -44,4 +52,14 @@ object WhichYearIsLossReported extends Enumerable.Implicits {
   implicit val enumerable: Enumerable[WhichYearIsLossReported] =
     Enumerable(values.map(v => v.toString -> v): _*)
 
+  private def generateYears(upTo: Int = 5): Seq[String] = {
```
lukastymo, line 55: I would probably use hardcoded enums, and update it every year. But if it is not doable I would implement dynamic function and put it in our common place where we have year-dependent code.

This function will probably go, but just for future: this function is pretty complex, I would unit test it, especially edge cases. To do this you need to make `private[adjustments]` so it is open in the same package and easily tested

lukastymo (changes requested):
I'm not convince 5 dynamic tax years should be implemented as enum

lukastymo (approved):

PR #418, opened 2024-08-07
[SASS-9316] Add logic for NICs Class 2 or Class 4 link

Jake-Raffe:


app/models/journeys/nics/TaxableProfitAndLoss.scala
```diff
@@ -0,0 +1,54 @@
...
+
+case class TaxableProfitAndLoss(taxableProfit: BigDecimal, taxableLoss: BigDecimal)
+
+object TaxableProfitAndLoss {
+  implicit val formats: Format[TaxableProfitAndLoss] = Json.format[TaxableProfitAndLoss]
+
+  def fromBusinessIncomeSourcesSummary(biss: BusinessIncomeSourcesSummary): TaxableProfitAndLoss =
+    TaxableProfitAndLoss(biss.taxableProfit, biss.taxableLoss)
+
+  def returnClassTwoOrFourEligible(taxableProfitsAndLosses: List[TaxableProfitAndLoss], userDoB: LocalDate, taxYear: TaxYear): NicClassExemption = {
```
Jake-Raffe, line 35: Separated this logic out of the view model so it can be reused for other pages

app/controllers/journeys/TaskListController.scala
```diff
@@ -52,24 +55,51 @@ class TaskListController @Inject() (override val messagesApi: MessagesApi,
...
-        message)
-    } yield Ok(view(taxYear, updatedRequest.user, tradeDetailsStatus, viewModelList, nationalInsuranceSummary)(updatedRequest, message))
+      updatedRequest     = taskListWithRequest.request
+      messages           = messagesApi.preferred(updatedRequest)
+      tradeDetailsStatus = getTradeDetailsStatus(taskListWithRequest)
+      completedTrades    = getTradesIfDetailsAreCompleted(taskListWithRequest.taskList)
+      businessSummaryList      <- getBusinessSummaries(completedTrades, updatedRequest, taxYear, messages)
+      nationalInsuranceSummary <- getNationalInsuranceSummary(taskListWithRequest, completedTrades, taxYear, messages)
+    } yield Ok(view(taxYear, updatedRequest.user, tradeDetailsStatus, businessSummaryList, nationalInsuranceSummary)(updatedRequest, messages))
     handleResultT(result)
```
Jake-Raffe, line 65: Tried to break up this huge for comprehension and make it a bit more descriptive

app/models/journeys/nics/NicClassExemption.scala
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.journeys.nics
+
+sealed trait NicClassExemption
+
+case object Class2      extends NicClassExemption
```
lukastymo, line 21: can you put it inside NicsClassExcemption object? Usually they live in companion objects

app/utils/TaxYearUtils.scala
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package utils
+
+import java.time.{LocalDate, Month}
+
+object TaxYearUtils {
```
lukastymo, line 21: We can use `TaxYear`'s companion object for utility methods for `TaxYear`

app/utils/TaxYearUtils.scala
```diff
@@ -0,0 +1,29 @@
...
+
+  val dateNow = LocalDate.now()
```
lukastymo, line 23: That looks like a val (you probably wanted def) otherwise it will be constant until we restart the app.

lukastymo (changes requested):
I'm not sure if the vals in TaxYearUtils were deliberately, also let's clarify if we want to use taxyear from the URL and inside TaxYear case class, or real current tax year (then it should probably go to companion  TaxYear object)

app/utils/TaxYearUtils.scala
```diff
@@ -0,0 +1,29 @@
...
+
+  val currentTaxYearStartDate =
```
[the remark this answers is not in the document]
Jake-Raffe: It replaces with the url taxyear in the ageIsBetween16AndStatePension method

lukastymo (approved):

PR #423, opened 2024-08-14
SASS-9279 - Net business loss - Claim a loss relief-create page

kbfield:


app/models/domain/BusinessIncomeSourcesSummary.scala
```diff
@@ -27,7 +29,11 @@ case class BusinessIncomeSourcesSummary(incomeSourceId: String,
                                         totalDeductions: Option[BigDecimal],
                                         accountingAdjustments: Option[BigDecimal],
                                         taxableProfit: BigDecimal,
-                                        taxableLoss: BigDecimal)
+                                        taxableLoss: BigDecimal) {
+  // TODO returnNetBusinessProfitForTaxPurposes logic is a placeholder, real value will be calculated in SASS-8626
+  def returnNetBusinessProfitForTaxPurposes(): BigDecimal = if (netLoss != 0) -netLoss else netProfit
```
lukastymo, line 34: we don't need if, `-0 == 0`, -netLoss is fine assuming != 0 is what you need

Can we have getNetBusinesProfitForTaxPurpose instead of return? get is more common

lukastymo (dismissed):

app/controllers/journeys/adjustments/profitOrLoss/CheckNetProfitLossController.scala
```diff
@@ -17,49 +17,61 @@
...
-        routes.CurrentYearLossesController.onPageLoad(taxYear, businessId, NormalMode)
+  def onPageLoad(taxYear: TaxYear, businessId: BusinessId): Action[AnyContent] = (identify andThen getData andThen requireData) async {
+    implicit request =>
+      val result = for {
+        incomeSummary <- service.getBusinessIncomeSourcesSummary(taxYear, request.nino, businessId, request.mtditid)
+        netAmount          = incomeSummary.returnNetBusinessProfitForTaxPurposes()
+        profitOrLoss       = incomeSummary.returnProfitOrLoss()
+        formattedNetAmount = formatSumMoneyNoNegative(List(netAmount))
+
+        table1 = buildTable1(profitOrLoss, 3000, 0.05, -3100)
```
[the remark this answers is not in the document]
Jake-Raffe: The issue is that table 2 and 3 flip their content depending on if it's a profit or loss so the title aren't consistent. Can be something that is refactored when the real values are fetched.
Can add it as a todo comment to consider renaming the tables in SASS-8626

app/models/domain/BusinessIncomeSourcesSummary.scala
Jake-Raffe: Business-wise, the value is called NetBusinessProfitForTaxPurposes and will be calculated from multiple values giving a positive value (profit) or negative value (loss).
As a placeholder we are using netLoss and netProfit (both are always positive BigDecimals) for stubbing purposes to test profit or loss on different businesses, with the opposite value (netProfit and netLoss respectively) set to zero.
So this is the method we'll use for now, but can change the name to `getNetBusinesProfitOrLossForTaxPurpose ` for clarity?

lukastymo: sounds good, I missed it is netProfit after else

app/controllers/journeys/adjustments/profitOrLoss/CheckNetProfitLossController.scala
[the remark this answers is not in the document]
lukastymo: ok

Jake-Raffe (approved):

PR #430, opened 2024-08-21
[SASS-9574] Update NICs Class 2 saving and loading from BE to handle Class2/4 answers model

Jake-Raffe:


app/models/journeys/nics/NICsJourneyAnswers.scala
```diff
@@ -0,0 +1,44 @@
...
+import play.api.libs.json.{Format, Json}
+import play.api.mvc.Result
+import play.api.mvc.Results.Redirect
+
+case class NICsJourneyAnswers(class2Answers: Option[NICsClass2Answers], class4Answers: Option[NICsClass4Answers])
+
+object NICsJourneyAnswers {
+  implicit val formats: Format[NICsJourneyAnswers] = Json.format[NICsJourneyAnswers]
+
+  def buildFromUserAnswers(userAnswers: UserAnswers): Either[Result, NICsJourneyAnswers] = {
```
lukastymo, line 32: I would add an unit test for this. Tests are also live documentation, and here we decided to fail when we have both class2 and class4 which would be nice to cover. 

Also this test would be nice safety net if we decide to mess with Class2 and Class4 mongo fields.

lukastymo (approved):

PR #440, opened 2024-09-03
[SASS-9015] Clear Simplified Expenses data when turnover is above threshold

Jake-Raffe:


app/controllers/journeys/income/IncomeCYAController.scala
```diff
@@ -76,9 +84,37 @@ class IncomeCYAController @Inject() (override val messagesApi: MessagesApi,
...
 
-      handleSubmitAnswersResult(context, result)
+    for {
+      existingTotalIncomeAmount    <- service.getTotalIncome(ctx)
+      currentAnswersTurnoverAmount <- EitherT.fromEither[Future](request.valueOrNotFoundError(TurnoverIncomeAmountPage, ctx.businessId))
+      currentAnswersOtherIncomeAmount = request.getValue[BigDecimal](OtherIncomeAmountPage, ctx.businessId).getOrElse(BigDecimal(0))
+      currentAnswersTotalIncomeAmount = currentAnswersTurnoverAmount + currentAnswersOtherIncomeAmount
+      existingIncomeIsOverThreshold   = totalIncomeIsOverIncomeThreshold(existingTotalIncomeAmount)
+      currentIncomeIsOverThreshold    = totalIncomeIsOverIncomeThreshold(currentAnswersTotalIncomeAmount)
+      expensesTailoringIsSimplified   = request.getValue(ExpensesCategoriesPage, ctx.businessId).exists(_ != IndividualCategories)
```
Jake-Raffe, line 112: This line uses an `.exists()` because if there are no expenses then there's no point trying to clear expenses from BE.
Will only be true if Expenses Categories == TotalAmount | NoExpenses

test/base/ControllerSpec.scala
```diff
@@ -79,6 +79,33 @@ trait ControllerTestScenarioSpec extends MockitoSugar with DefaultAwaitTimeout w
 
   }
 
+  case class TestStubbedScenario(userType: UserType = Individual,
```
Jake-Raffe, line 82: We can use this to phase out the mocking and use stubs instead in the BaseSpecs

PR #452, opened 2024-09-30
[SASS-10041] Income - Trading Allowance cap must include non-trading turnover

Jake-Raffe:


app/services/SelfEmploymentService.scala
```diff
@@ -255,9 +255,10 @@ object SelfEmploymentService {
   // TODO: Return an error as left once we have impl. error handling instead of redirecting.
   def getMaxTradingAllowance(businessId: BusinessId, userAnswers: UserAnswers): Either[Result, BigDecimal] =
     userAnswers
-      .get(TurnoverIncomeAmountPage, Some(businessId))
-      .fold(redirectJourneyRecovery().asLeft[BigDecimal]) { turnover =>
-        if (turnover > maxAllowance) maxAllowance.asRight else turnover.asRight
+      .get(TurnoverIncomeAmountPage, businessId)
+      .fold(redirectJourneyRecovery(Some("Turnover Income amount not found")).asLeft[BigDecimal]) { turnover =>
+        val totalAmount = turnover + userAnswers.get(NonTurnoverIncomeAmountPage, businessId).getOrElse(0)
+        if (totalAmount > maxAllowance) maxAllowance.asRight else totalAmount.asRight
```
Jake-Raffe, line 261: This is the change that is relevant to the ticket

app/controllers/journeys/income/NonTurnoverIncomeAmountController.scala
```diff
@@ -59,7 +59,10 @@ class NonTurnoverIncomeAmountController @Inject() (override val messagesApi: Mes
         .bindFromRequest()
         .fold(
           formWithErrors => Future.successful(BadRequest(view(formWithErrors, mode, request.userType, taxYear, businessId))),
-          answer => service.persistAnswerAndRedirect(page, businessId, request, answer, taxYear, mode)
+          answer => {
+            val updatedMode = if (request.getValue(page, businessId).contains(answer)) mode else NormalMode
+            service.persistAnswerAndRedirect(page, businessId, request, answer, taxYear, updatedMode)
```
Jake-Raffe, line 64: Currently, if I fill out Turnover and NonTurnover pages then I get the correct allowance cap on the final page. I can then go back and change either of these amount pages and return to CYA for submission without the allowance being updated. The cap or allowance amount might no longer be viable.

This change means that if the user changes their answer for either of these pages they must go through the rest of the journey, confirming their values with the updated allowance cap

Jake-Raffe:
Made some extra changes to condense similar methods from the controller or service object and moved them into an income package object.
Also, clear the data of some pages if the answer is changed so that the journey flow works properly with a dynamic trading allowance.

kbfield (approved):

PR #471, opened 2024-10-30
[SASS-10321] Bug - page submission should return to CYA if all journey pages are answered

Jake-Raffe:


app/viewmodels/journeys/adjustments/AdjustedTaxableProfitOrLossSummary.scala
```diff
@@ -76,6 +76,7 @@ object AdjustedTaxableProfitOrLossSummary {
       val netForTaxPurposesIsProfitOrLoss = if (amount < 0) Loss else Profit
       buildTableRow(s"profitOrLossCalculation.adjustedTable.netForTaxPurposes.$netForTaxPurposesIsProfitOrLoss", formattedAmount)
     }
+    val adjustedTaxableAmountIsProfitOrLoss = if (adjustedTaxableProfitOrLoss < 0) Loss else Profit
```
Jake-Raffe, line 79: This fixes an error where one of the lines on the adjusted profit or loss summary page has P / L content depending on if the journey is a net profit or loss. The content for this line is actually dependent on if the adjustedTaxableProfitOrLoss value is positive (P) or negative (L)

PR #476, opened 2024-11-05
SASS-10136R

ashwinirudrappa:
Returning customer / prior data for Balancing charge


app/controllers/actions/SubmittedDataRetrievalActionProvider.scala
```diff
@@ -76,7 +77,13 @@ class SubmittedDataRetrievalActionProviderImpl @Inject() (connector: SelfEmploym
         gtsouUpdated,
         Journey.CapitalAllowancesTailoring
       ) // TODO 10281 fetch Profit or Loss journey answers
```
Jake-Raffe, line 79: Not part of your ticket but could we fetch the Profit or Loss answers as well here since that has been missed previously

Jake-Raffe (dismissed):

Jake-Raffe (approved):

PR #483, opened 2024-11-20
[SASS-10383]Remove the Index endpoint

vfryer12:


app/controllers/standard/LanguageSwitchController.scala
```diff
@@ -28,7 +29,7 @@ class LanguageSwitchController @Inject() (
     cc: ControllerComponents
 ) extends LanguageController(languageUtils, cc) {
 
-  override def fallbackURL: String = routes.IndexController.onPageLoad.url
+  override def fallbackURL: String = appConfig.incomeTaxSubmissionStartUrl(TaxYear.dateNow.getYear)
```
palequezal, line 32: As we start to support more tax years, we can't necessarily assume that the user will want to be taken to the latest tax year; they might've been working on a previous year. There isn't really a neutral alternative place to land people right now though, so happy to keep this as it is for now, but we should take up with Team A & the UX team to see what we should do long term.

palequezal (approved):

PR #488, opened 2024-12-03
SASS-10371: Use the information of API#2085 to load the particular page

ashwinirudrappa:


app/controllers/journeys/adjustments/profitOrLoss/CurrentYearLossController.scala
```diff
@@ -44,51 +45,57 @@ class CurrentYearLossController @Inject() (override val messagesApi: MessagesApi
...
-      if (hasOtherIncomes) {
-        val filledForm = fillForm(whatDoYouWantToDoWithLossPage, businessId, formProvider(whatDoYouWantToDoWithLossPage, request.userType))
-        Ok(whatDoYouWantToDoWithLossView(filledForm, taxYear, businessId, request.userType, mode))
-      } else {
-        val filledForm = fillForm(carryLossForwardPage, businessId, booleanFormProvider(carryLossForwardPage, request.userType))
-        Ok(carryLossForwardView(filledForm, taxYear, businessId, request.userType, mode))
-      }
+      // TODO retrieve pension income in SASS-9566
+      val result = for {
+        hasOtherIncome <- service.hasOtherIncomeSources(taxYear, request.nino, request.mtditid)
```
Jake-Raffe, line 59: Since this is one line I'm not sure it needs to be in a for comprehension, it could just be a map

Jake-Raffe: Would be good to leave a placeholder for pension for clarity with the comment
```
val hasPensionIncome = true
val result = service.hasOtherIncomeSources(taxYear, request.nino, request.mtditid).map { hasOtherIncome =>
       if (hasOtherIncome && hasPensionIncome) {...
```

Jake-Raffe: same with onSubmit

app/controllers/journeys/adjustments/profitOrLoss/CurrentYearLossController.scala
```diff
@@ -44,51 +45,57 @@ class CurrentYearLossController @Inject() (override val messagesApi: MessagesApi
...
+          service.defaultHandleForm(
+            booleanFormProvider(carryLossForwardPage, request.userType),
+            carryLossForwardPage,
+            businessId,
+            taxYear,
+            toCyaIfAllPagesAnswered,
+            handleErrorFalse)
       }
+
+      handleApiResult(result).flatten
```
Jake-Raffe, line 98: can flatMap result on line 76 rather than map there and flatten here

it/connectors/SelfEmploymentConnectorISpec.scala
```diff
@@ -237,4 +238,31 @@ class SelfEmploymentConnectorISpec extends WiremockSpec with IntegrationBaseSpec
...
+    "return true when there is an other income" in {
+      val responseBody = "true"
+      stubGetWithResponseBody(checkForOtherIncomeSourcesUrl, OK, responseBody, headersSentToBE)
+
+      val result = connector.hasOtherIncomeSources(taxYear, nino, mtditid).value.futureValue
+
+      result shouldBe Right(true)
+    }
+
+    "return false on there is an other income" in {
```
Jake-Raffe, line 252: This description isn't very clear, or different from the 'true' case.
Could replace with "return 'true' from downstream" and "return 'false' from downstream".
But doesn't matter much, up to you

test/base/questionPages/CheckboxControllerBaseSpec.scala
```diff
@@ -60,7 +60,8 @@ abstract case class CheckboxControllerBaseSpec[A: Writes](controller: String, pa
   private def getRequest  = FakeRequest(GET, onPageLoadRoute)
   private def postRequest = FakeRequest(POST, onSubmitRoute).withFormUrlEncodedBody(("value", answer.toString))
 
-  def stubServiceSuccessfulSubmission = SelfEmploymentServiceStub(submitAnswerAndRedirectResult = Redirect(onwardRoute))
+  def stubServiceSuccessfulSubmission: SelfEmploymentServiceStub =
+    SelfEmploymentServiceStub(submitAnswerAndRedirectResult = Redirect(onwardRoute), hasOtherIncomeSources = Right(true))
```
Jake-Raffe, line 64: `hasOtherIncomeSources = Right(true)` is something you should override the stub in the ControllerSpec rather than the BaseSpec since it is only relevant to this one Controller, not all CheckboxControllers

test/controllers/journeys/adjustments/profitOrLoss/CurrentYearLossControllerCarryLossForwardSpec.scala
```diff
@@ -42,7 +45,7 @@ class CurrentYearLossControllerCarryLossForwardSpec extends BooleanGetAndPostQue
     view(expectedForm, taxYear, businessId, scenario.userType, scenario.mode).toString()
   }
 
-  // TODO in SASS-9566 test that the view CarryLossForward gets displayed when user has NO other incomes
-}
```
Jake-Raffe, line 46: Should a comment remain to say to test for when user has no otherIncomeSources but has pensionIncomes (to match the change in the controller)?

test/controllers/journeys/adjustments/profitOrLoss/CurrentYearLossControllerSpec.scala
```diff
@@ -0,0 +1,133 @@
...
+import play.api.mvc.Results.Redirect
+import play.api.mvc.{Call, Request}
+import play.api.test.FakeRequest
+import play.api.test.Helpers._
+import stubs.services.SelfEmploymentServiceStub
+import views.html.journeys.adjustments.profitOrLoss.WhatDoYouWantToDoWithLossView
+
+class CurrentYearLossControllerSpec extends ControllerSpec {
+
+  def answer: Set[WhatDoYouWantToDoWithLoss] = Set(DeductFromOtherTypes)
```
Jake-Raffe, line 43: Is this not supposed to use the CheckboxControllerBaseSpec that you edited in this PR for this purpose?

wp94 (approved):

PR #489, opened 2024-12-04
[SASS-10588] Add EMA Supporting Agent FS and authorisation logic

MJCallahanPage:
Notes:
- Adds Feature Switch for the Supporting/Secondary Agent authorisation
- Logic added to Agent authentication when Feature Switch is enabled
- Updates the `User` class to have a flag for `isSupportingAgent` which may be necessary to limit functionality as part of future stories
- Refactors the AuthSpec so that Scalamock is utilised to simulate different responses from AuthConnector to make tests more verbose, better coverage and specific scenarios tested

App-Config-* PR(s) to be reviewed at the same time:
- https://github.com/hmrc/app-config-base/pull/11376

### DIFF
--- a/app/config/FrontendAppConfig.scala
+++ b/app/config/FrontendAppConfig.scala
@@ -26,36 +26,37 @@ import uk.gov.hmrc.play.bootstrap.config.ServicesConfig
 @Singleton
 class FrontendAppConfig @Inject() (configuration: Configuration, servicesConfig: ServicesConfig) {
 
-  val host: String    = configuration.get[String]("host")
-  val appName: String = configuration.get[String]("appName")
+  lazy val host: String    = configuration.get[String]("host")
+  lazy val appName: String = configuration.get[String]("appName")
 
-  private val contactHost                  = configuration.get[String]("contact-frontend.host")
-  private val contactFormServiceIdentifier = "income-tax-self-employment-frontend"
+  private lazy val contactHost                  = configuration.get[String]("contact-frontend.host")
+  private lazy val contactFormServiceIdentifier = "income-tax-self-employment-frontend"
 
   def feedbackUrl(implicit request: RequestHeader): String =
     url"$contactHost/contact/beta-feedback?service=$contactFormServiceIdentifier&backUrl=${host + request.uri}".toString
 
-  val loginUrl: String         = configuration.get[String]("urls.login")
-  val loginContinueUrl: String = configuration.get[String]("urls.loginContinue")
-  val signOutUrl: String       = configuration.get[String]("urls.signOut")
+  def loginUrl: String         = configuration.get[String]("urls.login")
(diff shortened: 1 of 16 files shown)


app/config/FrontendAppConfig.scala
```diff
@@ -26,36 +26,37 @@ import uk.gov.hmrc.play.bootstrap.config.ServicesConfig
 @Singleton
 class FrontendAppConfig @Inject() (configuration: Configuration, servicesConfig: ServicesConfig) {
 
-  val host: String    = configuration.get[String]("host")
-  val appName: String = configuration.get[String]("appName")
+  lazy val host: String    = configuration.get[String]("host")
```
basotia, line 29: I believe we can leave them as `val` They are unlikely to change, also making them lazy might lead to runtime issues if configs are missing.

MJCallahanPage: They have to be lazy for aid mocking. But the alternative would be to create a trait and then a module binding for it to a concreate class. So, I can update this to follow that pattern instead and then they can switch back to non-lazy.

I'll refactor this now 👍

app/config/FrontendAppConfig.scala
```diff
@@ -26,36 +26,37 @@ import uk.gov.hmrc.play.bootstrap.config.ServicesConfig
...
+  private lazy val contactHost                  = configuration.get[String]("contact-frontend.host")
+  private lazy val contactFormServiceIdentifier = "income-tax-self-employment-frontend"
 
   def feedbackUrl(implicit request: RequestHeader): String =
     url"$contactHost/contact/beta-feedback?service=$contactFormServiceIdentifier&backUrl=${host + request.uri}".toString
 
-  val loginUrl: String         = configuration.get[String]("urls.login")
-  val loginContinueUrl: String = configuration.get[String]("urls.loginContinue")
-  val signOutUrl: String       = configuration.get[String]("urls.signOut")
+  def loginUrl: String         = configuration.get[String]("urls.login")
```
basotia, line 38: Please correct me if i have overlooked something here, was there a reason we changed it to def?

app/config/FrontendAppConfig.scala
```diff
@@ -26,36 +26,37 @@ import uk.gov.hmrc.play.bootstrap.config.ServicesConfig
...
+  def loginContinueUrl: String = configuration.get[String]("urls.loginContinue")
```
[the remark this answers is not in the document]
MJCallahanPage: Same as above one :)

app/config/FrontendAppConfig.scala
MJCallahanPage: It's so that it can be mocked with `scalamock` - as you can't mock a val with that. So this has been refactored now so that there is a `trait` with a `def` and these can then override with concrete `val` still:
https://scalamock.org/user-guide/faq/#can-i-mock-val--lazy-val

basotia (approved):

PR #494, opened 2025-01-16
Consider correct tax year for age calculation

ashwinirudrappa:


ashwinirudrappa:
test this please

palequezal:
As discussed on Slack, I'm not comfortable with us changing the date logic here without understanding the ramifications. Given we know the issue is with the tests and not the code, set the few failing tests to ignore and we'll fix the tests properly at the end of the sprint to avoid this taking too much of our time.

wp94 (approved):

PR #545, opened 2025-04-11
SASS-11363- 15. Travel and accommodation journey - Create "Check your answers" page

iyadi:


app/models/common/JourneyStatus.scala
```diff
@@ -39,7 +39,8 @@ object JourneyStatus extends Enum[JourneyStatus] with utils.PlayJsonEnum[Journey
     status // If backend returns NotStarted -> answers are submitted but status not submitted -> UI journey status is InProgress
       .collect {
         case NotStarted => InProgress
-        case status     => status
+        case statos =>
```
wp94, line 42: typo in the word status

iyadi: this has been done in purpose 
a spelling mistake to distinguish it from 
`val status: Option[JourneyStatus]`
it was given a compiler warning.
in fact this call does not need to be collect it is a map, and collect gives the wrong meaning that it is filtering the option but it is not fact.

wp94 (approved):

PR #549, opened 2025-04-16
SASS-11452R: Make screens 3-9 indexable

ashwinirudrappa:


app/models/Index.scala
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package models
+
+import play.api.libs.json.{Format, Json}
+import play.api.mvc.PathBindable
+
+case class Index(value: Int) {
+
+  // def position: Int = display - 1
```
wp94, line 24: Is this commented out line kept on purpose?

wp94 (approved):

PR #550, opened 2025-04-22
Improve the unit tests performance

ashwinirudrappa:


ashwinirudrappa:
test this please

test/base/questionPages/BooleanGetAndPostQuestionBaseSpec.scala
```diff
@@ -74,35 +76,32 @@ abstract case class BooleanGetAndPostQuestionBaseSpec(controllerName: String, pa
...
-            running(application) {
-              val result = route(application, getRequest).value
+            val result: Future[Result] = route(application, getRequest).value
 
-              status(result) shouldBe OK
-              contentAsString(result) shouldBe expectedView(form().fill(validAnswer), this)(getRequest, messages(application), application)
-            }
+            status(result) shouldBe OK
+            contentAsString(result) shouldBe expectedView(form().fill(validAnswer), this)(getRequest, messages(application), application)
+            application.stop()
```
palequezal, line 83: So this works, but it would be onerous to have to add this to every test. I think the better solution would be to make `SpecBase` extend `GuiceOneAppPerTest`, which will do all that set up/tear down automatically, and the app can be overridden in the spec to add in config and extra modules as needed.

Ideally, we'd use `GuiceOneAppPerSuite` and that would go even faster because less apps would be built.

PR #559, opened 2025-05-15
Sass 11527 b

ellie-l-newell:


conf/messages.en
```diff
@@ -2529,4 +2530,4 @@ literaryOrCreativeWorks.p.agent = This includes literary, dramatic, musical or a
 literaryOrCreativeWorks.checkYourAnswersLabel = literaryOrCreativeWorksLabel
 literaryOrCreativeWorks.error.required.individual = Select yes if you got over half of your profit from literary or creative works
 literaryOrCreativeWorks.error.required.agent = Select yes if your client got over half their profit from literary or creative works
-literaryOrCreativeWorks.change.hidden = LiteraryOrCreativeWorks
+literaryOrCreativeWorks.change.hidden = Change if you got over half of your profit from literary or creative works
```
russellthorn, line 2533: Is the word change needed here - i.e. is the visual change link set to aria-hidden=true or not. I would recommend that it is not and the/this change in the hidden text is not not as per the change link/CYA pattern https://design-system.service.gov.uk/patterns/check-answers/

ashwinirudrappa (approved):

PR #569, opened 2025-06-06
[SASS-11696] Add SessionDataService and update IdentifierAction to call it

MJCallahanPage:
Includes some minor refactoring to tidy up a few things and reduce some complexity in the tests. (two duplicate test specs have been deleted because the AuthAction spec covers all the Auth testing now).


matthewwardcap (changes requested):
Could you maybe update test.sh to do a check for coverage? Though honestly I can't get a full successful run due to the open files issue caused by the scaffold generated tests, so might be a bit pointless

Following could do with a bump:
bootstrap - 9.13.0
enumeratum - 1.9.0
enumeratum-play-json - 1.9.0
play-frontend-hmrc-play-30 - 12.5.0

